PR #60, opened 2026-02-21
Fix iCal adapter: detect non-ICS responses (HTML from deactivated plugins)

johnrclem:
The BAH3 scraper fetches 128KB of data but finds 0 VEVENTs because the
server is returning HTML instead of ICS content — likely because the
All-in-One Event Calendar WordPress plugin was deactivated or removed.

Previously, node-ical would silently parse the HTML and return an empty
result with no errors. Now the adapter validates that the response body
starts with BEGIN:VCALENDAR before parsing, and returns a clear error
message with content-type and body preview when the response is not
valid ICS.

Also adds contentType to diagnosticContext for all responses.

https://claude.ai/code/session_013GUWkzg8ekCGYqancywTbk


johnrclem:
@gemini-code-assist review

PR #202, opened 2026-03-08
fix: double-header matching for single existing event

johnrclem:
## Summary
- **Fix single-event matching**: When only one canonical Event exists for a kennel+date, no longer blindly matches it. If both incoming and existing events have `sourceUrl` and they differ, creates a new event (double-header detection).
- **Strict disambiguation cascade**: Changed sequential `if` to `else if` for multi-event matching — if `sourceUrl` is present but doesn't match, don't fall through to startTime/title.
- **Reconcile key includes `startTime`**: Prevents two events with same sourceUrl but different startTime from collapsing during stale-event reconciliation.

Addresses PR #201 review feedback (Gemini Code Assist) and fixes the root cause where the second Boston H3 event on 2026-03-08 was overwriting the first instead of creating a separate canonical Event.

## Test plan
- [x] All 2264 tests pass (106 test files)
- [x] Updated "creates second event" test to expect 2 creates
- [x] Added "creates new when single existing has different sourceUrl" test
- [x] Added "matches single existing when incoming has no sourceUrl" backward-compat test
- [x] Updated reconcile test mocks with `startTime` field
- [ ] Reset mis-linked RawEvent `cmmieiej5004704kwlvrw7fvw` (`processed: false, eventId: null`)
- [ ] Re-scrape Boston Hash Calendar → verify both events appear as separate canonical Events
- [ ] Re-scrape again → verify no duplicates (dedup still works)

### DIFF
--- a/src/pipeline/merge.test.ts
+++ b/src/pipeline/merge.test.ts
@@ -393,20 +393,44 @@ describe("double-header support", () => {
     expect(result.created).toBe(0);
   });
 
-  it("creates new event when single existing has different sourceUrl", async () => {
+  it("cross-source: matches single existing event with different sourceUrl and creates EventLink", async () => {
     mockRawEventFind.mockResolvedValueOnce(null);
     mockEventFindMany.mockResolvedValueOnce([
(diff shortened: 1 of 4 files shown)


johnrclem:
@codex review

PR #282, opened 2026-03-21
fix: Strava suggestions — scoring fixes, UI redesign, code cleanup

johnrclem:
## Summary

Addresses 13 production feedback issues with the Strava suggestions feature from PR #280.

### Scoring fixes (backend)
- Filter non-running activities (yoga, cycling) and <1km distances from suggestions
- Fix geo scoring: default 0 when coords missing (was 0.2 — caused cross-country matches)
- Add nameScore gate: generic names like "Afternoon Run" no longer match everything
- Raise match threshold from 0.5 → 2.0 (requires meaningful signals)
- Sort by confidence score (not date), cap at 10 (was 30)
- Add `matchReasons` array explaining why each match was suggested

### UI redesign (frontend)
- Event-first card layout: kennel name as primary label (blue, clickable), Strava activity as secondary
- Match reason tags: "Same day" · "Within 3 km" · "Name: NYC H3"
- Clickable links: kennel → event page, activity name → Strava
- Three actions: "I Was There" / "Not a Hash" / "Skip" (client-side only)
- Sync status + Sync button in header (was buried in footer)
- Activity dates shown in Link section
- Section subtitles for clarity ("Hash events that match..." / "Attach Strava data to...")

### Code cleanup (simplify review)
- `scoreMatch()` returns `ScoreBreakdown` struct — eliminates double-computation of haversineDistance/fuzzyNameMatch
- Export `timeToMinutes` from match-score.ts, remove duplicate
- Hoist `SCOREABLE_SPORTS` to module-level constant (was inside `.map()` loop)
- Replace inline `formatTimeAgo` with existing `formatRelativeTime`

## Test plan
- [ ] TypeScript compiles clean
- [ ] 99 strava + logbook tests pass (including 2 new geo-scoring tests)
- [ ] Only running activities appear in suggestions (no yoga/cycling)
- [ ] Geo matches are geographically relevant (no cross-country)
- [ ] Event is primary label, Strava activity is secondary
- [ ] Match reason tags visible on cards
- [ ] Skip vs Not a Hash behavior differs (skip = temporary, not a hash = permanent)
- [ ] Sync status and button visible in section header

### DIFF
--- a/src/app/strava/actions.ts
+++ b/src/app/strava/actions.ts
@@ -394,12 +394,14 @@ export async function getUnmatchedStravaActivities(): Promise<
     const earliest = new Date(Math.min(...attDates));
     const latest = new Date(Math.max(...attDates));
 
-    // Get unmatched, non-dismissed Strava activities within date range
+    // Get unmatched, non-dismissed running Strava activities within date range
     const activities = await prisma.stravaActivity.findMany({
       where: {
         stravaConnectionId: connection.id,
         matchedAttendanceId: null,
         matchDismissed: false,
+        sportType: { in: SCOREABLE_SPORTS_ARRAY },
+        distanceMeters: { gte: 1000 },
         dateLocal: {
           gte: earliest.toISOString().substring(0, 10),
           lte: latest.toISOString().substring(0, 10),
--- a/src/components/logbook/StravaSuggestions.tsx
+++ b/src/components/logbook/StravaSuggestions.tsx
@@ -482,9 +482,9 @@ function SuggestionCard({
         {/* Line 2: Match reasons */}
         {reasons.length > 0 && (
           <div className="flex gap-1 flex-wrap">
-            {reasons.map((reason, i) => (
+            {reasons.map((reason) => (
               <span
-                key={i}
+                key={reason}
                 className={`text-[10px] px-1.5 py-0.5 rounded font-mono ${
                   reason.startsWith("Same") || reason.startsWith("Within")
                     ? "bg-emerald-50 text-emerald-700"
--- a/src/lib/strava/match-score.test.ts
+++ b/src/lib/strava/match-score.test.ts
@@ -52,13 +52,15 @@ describe("scoreMatch", () => {
     expect(score.total).toBeGreaterThan(0);
   });
 
-  it("returns 0 for generic activity names (nameScore gate)", () => {
+  it("scores low for generic activity names (filtered by threshold in suggestions)", () => {
     const score = scoreMatch(
       { activityName: "Morning Run", stravaSportType: "Run", stravaTimeLocal: null },
       "BH3",
       null,
     );
-    expect(score.total).toBe(0);
+    // Generic names still score (for findBestMatchIndex) but below suggestion threshold (2.0)
+    expect(score.total).toBeLessThan(2.0);
+    expect(score.total).toBeGreaterThan(0);
   });
 
   it("returns breakdown with geoKm when coords provided", () => {
--- a/src/lib/strava/match-score.ts
+++ b/src/lib/strava/match-score.ts
@@ -53,8 +53,9 @@ export function scoreMatch(
   // 1. Name match (0–1, weighted 3x)
   const nameScore = fuzzyNameMatch(activity.activityName, kennelShortName);
 
-  // Early exit: generic names like "Afternoon Run" should not match anything
-  if (nameScore < 0.15) return ZERO_BREAKDOWN;
+  // Low name scores get scored but won't pass the suggestion threshold (2.0).
+  // We still compute the full breakdown so findBestMatchIndex can rank by time/geo
+  // even when all candidates have generic names like "Afternoon Run".
 
   // 2. Geo proximity (0–1, weighted 2x)
   let geoScore = 0; // no coords = no credit


johnrclem:
## Feedback: Hover Text Consistency

### Bug: Missing tooltip on kennel short name in `StravaSuggestions.tsx`

The `SuggestionCard` renders the kennel short name as a plain `<a>` link without a tooltip showing the full kennel name:

```tsx
<a href={`/hareline/${s.eventId}`} className="font-semibold text-sm text-blue-500 hover:underline">
  {s.kennelShortName}
</a>
```

By contrast, `LogbookList.tsx` wraps the kennel short name in a `<Tooltip>` → `<TooltipContent>{entry.event.kennel.fullName}</TooltipContent>` pattern. The same treatment should be applied here to match the logbook row experience.

### Bug: Missing tooltip on region badge in `StravaSuggestions.tsx`

The region badge in `SuggestionCard` is a raw `<span>` with only a `title` attribute (native browser tooltip):

```tsx
<span className={`... ${colorCls}`} title={s.kennelRegion}>
  {abbrev}
</span>
```

The shared `<RegionBadge>` component already wraps the abbreviation in a `<Tooltip>` + `<TooltipContent>` and should be used here instead of the inline span, just as `LogbookList.tsx` does. Same issue applies to `QuickAddDialog.tsx`'s `EventRow`, which also renders a raw `<span>` with no tooltip for the region abbreviation.

### Bug: Missing tooltip on kennel short name in `page.tsx` (homepage)

The homepage "Coming Up" event cards render the kennel short name as plain text with no tooltip:

```tsx
<span className="text-base font-extrabold tracking-tight text-foreground">
  {event.kennel.shortName}
</span>
```

The `fullName` is fetched in the query but never surfaced to the user. A tooltip showing `event.kennel.fullName` on hover would make this consistent with the logbook view.

### Summary

Three locations need the same tooltip treatment that `LogbookList.tsx` already implements correctly:

| Location | Element | Fix needed |
|---|---|---|
| `StravaSuggestions.tsx` `SuggestionCard` | Kennel short name `<a>` | Wrap in `<Tooltip>` showing `fullName` |
| `StravaSuggestions.tsx` `SuggestionCard` | Region badge `<span>` | Replace with `<RegionBadge>` component |
| `QuickAddDialog.tsx` `EventRow` | Region badge `<span>` | Replace with `<RegionBadge>` component |
| `page.tsx` (homepage) | Kennel short name `<span>` | Wrap in `<Tooltip>` showing `fullName` (requires passing `fullName` through to JSX) |

Note: `RegionBadge` already has both `title={region}` and `<TooltipContent>{region}</TooltipContent>` built in, so switching to it covers both the accessibility `title` and the styled tooltip in one step.

johnrclem:
## Bug: Geo-scoring fails to filter geographically impossible suggestions

The suggestions list currently shows hash events in **London (BarnesH3/LDN)**, **Dayton (Dayton H4/DAY)**, **Georgia (HVH3/PV)**, and **MAC** region against a Strava activity that is clearly geo-tagged to New York City. These suggestions are not just unhelpful — they actively undermine trust in the feature.

### Root cause

The scoring formula in `match-score.ts` is:

```
total = nameScore×3 + geoScore×2 + timeScore×1 + sportBonus×0.2
```

`geoScore` is only non-zero when **both** the Strava activity **and** the event have coordinates. When event coordinates are missing — which is common for many kennels — `geoScore` defaults to `0` and the match proceeds on time alone. A `timeScore` of ~0.5–0.8 ("Similar time") plus `sportBonus` of 0.2 is enough to clear the 2.0 threshold when combined with even a partial `nameScore`, producing cross-continent suggestions.

The fix from PR #282 ("default `geoScore` to 0 when coords missing, was 0.2") only prevents a _false positive_ geo boost — it does **not** prevent a _false negative_ geo exclusion. If the Strava activity has GPS coords but the event does not, geo provides zero signal either way and the suggestion passes through on time alone regardless of continent.

### Evidence from production

Live logbook with a Strava run clearly in NYC shows:
- **BarnesH3** (London, LDN) — "Same day · Similar time" — no geo tag on event
- - **HVH3** (PV/Piedmont Virginia) — "Same day · Similar time"
- - **W3H3** (MAC region) — "Same day · Similar time" — despite activity labeled "NYC H3 · 7.0 mi"
- - **Dayton H4** (DAY, Ohio) — "Same day · Similar time · Afternoon Run (10.3 mi)"
None show a geo match reason tag, confirming event-side coords are absent for all four.

### Proposed fix

When the Strava activity **has** GPS coordinates but the matched event does **not**, geo should act as a **disqualifier** rather than a no-op. Options in order of preference:

1. **Hard disqualifier** — if `activity.startLat != null` and `eventLat == null`, skip the candidate (or apply `geoScore = -1` so total cannot reach the 2.0 threshold without a very strong name match)
2. 2. **Raise per-candidate threshold** — when geo is unavailable, require `nameScore > 0.5` as a prerequisite rather than just the global total threshold
3. 3. **Soft penalty** — `geoScore = -0.5` when activity has coords but event doesn't, making it very hard to reach 2.0 on time + name alone
Option 1 is cleanest: if we know the user ran in NYC, only suggest events that either (a) have coordinates near NYC, or (b) have no coordinates at all **and** a strong name/time match.

### Also affects "Link Strava to check-ins"

`getUnmatchedStravaActivities` doesn't use `scoreMatch` at all — it pairs same-day activities to same-day attendances with zero geo filtering or threshold. The "Link Strava to check-ins" section has the same cross-continent matching problem with no protection whatsoever.

johnrclem:
## Follow-up: Geo penalty insufficient — cross-continent suggestions still appearing (post-#285)

After the `-0.25` geo penalty in #285, **BARH3 (San Francisco)** and **Dayton H4** are still being suggested for runs on the East Coast. Here's the exact math on why.

### Why the penalty isn't working

The penalty only fires when `activity.startLat != null && eventLat == null`. But Strava does not always return GPS start coordinates — for indoor runs, treadmill activities, or manually-entered activities, `startLat`/`startLng` are null and stored as null in `StravaActivity`. When both sides have null coords, `geoScore = 0` (no penalty, no boost), and the suggestion passes on time + name alone.

### Failing cases with exact scores (threshold = 2.0)

**BARH3 (SF) vs "NYC H3" activity** — activity has no stored GPS coords, event has no coords:
```
nameScore = fuzzy("NYC H3", "BARH3") = 0.333  →  ×3 = 1.000
geoScore  = 0  (both null)                    →  ×2 = 0.000
timeScore = 1 - |18:25 - ~18:30| / 120 ≈ 0.958  →  ×1 = 0.958
sportBonus = 0.2
total = 2.158  ✗ passes threshold
```

**Dayton H4 vs "Afternoon Run"** — activity has no stored GPS coords, event has no coords:
```
nameScore = fuzzy("Afternoon Run", "Dayton H4") = 0.308  →  ×3 = 0.923
geoScore  = 0  (both null)                               →  ×2 = 0.000
timeScore = 1 - |14:01 - 14:00| / 120 ≈ 0.992            →  ×1 = 0.992
sportBonus = 0.2
total = 2.115  ✗ passes threshold
```

**HVH3 (PV) vs "Sav H3"** — activity has no stored GPS coords, event has no coords:
```
nameScore = fuzzy("Sav H3", "HVH3") = 0.500  →  ×3 = 1.500
geoScore  = 0  (both null)                   →  ×2 = 0.000
timeScore = 0.3 (default, no event startTime) →  ×1 = 0.300
sportBonus = 0.2
total = 2.000  ✗ passes threshold exactly (condition is `< 2.0`, not `<= 2.0`)
```

### Root problems

**1. `nameScore` is too loose for short H3 names.** `fuzzy("Sav H3", "HVH3")` = 0.500 because Levenshtein on short strings like these produces high similarity. Any 4–6 character kennel code has a high chance of scoring ≥ 0.4 against any 6-character activity name purely by character overlap. The threshold of 2.0 is too easy to clear when `nameScore * 3` contributes ~1.5 from a meaningless partial match.

**2. "Afternoon Run" is a generic name that should score 0.** The `nameScore` gate that was present in an earlier version (blocking scores when the name is generic) was removed as part of the PR #282 cleanup. `fuzzy("Afternoon Run", "Dayton H4")` = 0.308, which multiplied by 3 = 0.923, nearly enough to clear threshold with a good time match alone. Generic activity names (those containing "Run", "Walk", "Ride", "Morning", "Afternoon", "Evening") should be gated back to a lower weight or zero nameScore — these names carry zero kennel-identity signal.

**3. The `-0.25` penalty doesn't cover the null/null case.** When neither side has coordinates, there is no geographic signal at all. The penalty was designed to protect against "activity has GPS, event doesn't" — but the common case in production is "neither has GPS" and it gets a free pass.

### Proposed fixes

1. **Raise the threshold to 2.5** when neither side has coordinates — this blocks all three failing cases above while preserving legitimate matches with stronger name or time signals.
2. 2. **Zero out `nameScore` for generic activity names** (`/^(morning|afternoon|evening|lunch|night)?\s*(run|walk|ride|hike|workout)s?$/i`) — these names should not contribute positively to any match.
3. 3. **Raise the minimum `nameScore` gate** — require `nameScore >= 0.4` (not just total >= 2.0) before a no-geo match can pass. `fuzzy("Afternoon Run", "Dayton H4")` = 0.308 would be blocked; `fuzzy("Sav H3", "HVH3")` = 0.500 would still require review.
### Open item from previous feedback not yet addressed

The `getUnmatchedStravaActivities` function ("Link Strava to check-ins" section) still does **zero scoring** — it pairs by same day only with no `scoreMatch` call, no threshold, and no geo filtering. Every case documented above also applies there with no protection at all.

PR #476, opened 2026-04-06
chore(deploy): drop --accept-data-loss from vercel-build

johnrclem:
## Why

Today's \`AuditLog\` table got silently wiped because the Vercel build runs \`prisma db push --accept-data-loss\` on every deploy. For most schema diffs that's fine (column drops preserve rows), but for some diffs Prisma chooses to **recreate** the table — which nukes all rows without warning. We hit one of those today and lost all Phase 1+2 audit history.

## What

One-line script change + a CLAUDE.md note documenting the new workflow.

\`\`\`diff
- "vercel-build": "prisma db push --accept-data-loss && next build",
+ "vercel-build": "prisma db push && next build",
\`\`\`

After this lands, any PR whose schema change requires destructive DB ops will fail the Vercel deploy with a Prisma data-loss error. The fix workflow becomes:

1. Author a manual SQL file in \`prisma/manual-sql/\` for the destructive change
2. Run it against the live DB before merging the PR
3. Merge — the converged schema/DB lets \`db push\` proceed without the flag

Same pattern we already use for partial unique indexes, GCAL frequency bumps, and audit suppressions.

## Test plan
- [x] Diff confirmed minimal — 1 logic line + 1 doc paragraph
- [x] Referenced example SQL files exist in \`prisma/manual-sql/\`
- [ ] Vercel preview deploy succeeds (no pending destructive ops)
- [ ] Future destructive PR fails fast with a clear error (this is the regression test by definition)

## Follow-up
Filing a separate issue for the longer-term move to versioned migrations (\`prisma migrate deploy\`) — it's the proper fix but a bigger lift.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -509,7 +509,15 @@ Large reference lists are in `.claude/rules/`. They load automatically when you 
 - `rules/live-verification.md` — Mandatory live adapter verification (loads for adapter work)
 
 ## Schema Changes (Destructive Ops)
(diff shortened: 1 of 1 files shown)


johnrclem:
@codex @claude review

PR #527, opened 2026-04-08
fix: treat Hash Rego late-night times as placeholders (#487)

### LINKED ISSUE #487: [Audit] EWH3 — Event Quality [event-improbable-time] (1 events) — 2026-04-07
Automated audit found **1 event** affected by `event-improbable-time` for **EWH3**.

**Severity:** warning
**Adapter:** HASHREGO
**Rule:** `event-improbable-time`

## Sample Events

- **https://www.hashtracks.xyz/hareline/cmlrh0fms003e04jxcrz52a6g**
  - Current: `"23:45"`
  - Source: https://www.google.com/calendar/event?eid=OTZwbDMxN2M2c25iMDgzMjQ5Ym4zdWVwbWZfMjAyNjA0MDkgZXdoM2hhcmVyYXpvckBt

## Fix Guidance

Start time is between 23:00–04:00, which is unusual for a hash run. Check if this is a timezone conversion issue (UTC stored as local) or genuinely a late-night/early-morning event.

johnrclem:
## Summary

Closes #487. The audit flagged EWH3 1355 (\"National Unicorn Day\") with a start time of \`23:45\` on 2026-04-09. Investigation shows hashrego.com literally renders \`11:45 PM\` in the index row for that event — the kennel mismanagement used \`11:45 PM\` as their \"no time set\" placeholder instead of the more common \`11:59 PM\` that the existing parser already handled.

Real hash runs don't start between 11 PM and 4 AM on the Hash Rego platform, so the whole band is now treated as absent. \`parseHashRegoTime\` returns \`null\` for any time in \`23:00–03:59\` instead of only the exact \`11:59 PM\` string.

## Why not a narrower fix

I considered adding \`11:45 PM\` to an allowlist, but that pushes the problem down the road — the next placeholder variant (e.g., \`11:30 PM\`, \`12:15 AM\`) would re-raise the same bug. The late-night band is a more robust rule that matches reality: the \`event-improbable-time\` audit rule itself already defines this same window as \"unusual for a hash run\", so the parser is now consistent with what the auditor would flag anyway.

## Tests

- New cases for \`11:45 PM\` (the #487 variant), the band endpoints (\`11:00 PM\`, \`3:59 AM\`), and just-outside values (\`10:59 PM\`, \`4:00 AM\`)
- Dropped the \`12:00 AM → 00:00\` test — midnight is equally implausible for a real hash run on this platform and is now part of the band

\`npx vitest run src/adapters/hashrego\` → 95/95 ✅
\`npm test\` → 3989 passing ✅
\`npx tsc --noEmit\` ✅

## Test plan

- [x] Unit tests for the new placeholder band
- [x] Full suite passes
- [ ] After merge: trigger the Hash Rego scrape; EWH3 1355 RawEvent should have \`startTime: undefined\` instead of \`\"23:45\"\`
- [ ] Verify #487 auto-closes via verify-fixes workflow

### DIFF
--- a/src/adapters/hashrego/adapter.test.ts
+++ b/src/adapters/hashrego/adapter.test.ts
@@ -78,8 +78,15 @@ describe("parseHashRegoTime", () => {
     expect(parseHashRegoTime("11:45 PM")).toBeNull();
   });
 
-  it("returns null for 11:00 PM and 3:59 AM (late-night band endpoints)", () => {
+  it("returns null for 11:00 PM (late-night band PM endpoint)", () => {
     expect(parseHashRegoTime("11:00 PM")).toBeNull();
+  });
+
+  it("returns null for 1:00 AM (mid-band AM time)", () => {
+    expect(parseHashRegoTime("01:00 AM")).toBeNull();
+  });
+
+  it("returns null for 3:59 AM (late-night band AM endpoint)", () => {
     expect(parseHashRegoTime("03:59 AM")).toBeNull();
   });
 


johnrclem:
@claude review

PR #603, opened 2026-04-11
Make MATCHED a semi-terminal discovery status (fix slug degradation)

johnrclem:
## Summary

- After the PR #554 backfill restored 205 SourceKennel rows for the HASHREGO source, the count dropped back to ~12 within a day — twice (Apr 8→9 and again after PR #593 merge).
- Root cause: `syncKennelDiscovery` re-evaluated MATCHED discoveries on every run. Fuzzy scores fluctuate near the 0.95 threshold, flipping MATCHED↔NEW and triggering the downgrade cleanup block which deleted the SourceKennel row each cycle. 193 of 205 rows cycled this way.
- Fix: add MATCHED to the `isTerminal` set so subsequent syncs skip re-scoring and take the `updateTerminalDiscovery` path (refreshes profile data without touching match state). Only explicit admin actions can change status. Remove the now-dead downgrade cleanup block.

## Test plan
- [x] `npm run lint` clean (pre-existing warnings only)
- [x] `npm test` — 4305 passing. Replaced downgrade/retarget tests with a "MATCHED is semi-terminal" test asserting no re-scoring + no SourceKennel deletion + profile still refreshed.
- [ ] Re-run backfill: `BACKFILL_ALLOW_SELF_SIGNED_CERT=1 npx tsx scripts/backfill-hashrego-source-kennels.ts --apply`. Confirm SourceKennel count returns to ~205.
- [ ] Run `/admin/discovery → Sync Now` twice in a row. Confirm SourceKennel count stays at ~205 after both runs.
- [ ] Trigger a HASHREGO scrape and confirm `kennelSlugsConfigured.length ≈ 205`.

### DIFF
--- a/src/pipeline/kennel-discovery.test.ts
+++ b/src/pipeline/kennel-discovery.test.ts
@@ -202,9 +202,15 @@ describe("syncKennelDiscovery", () => {
     const result = await syncKennelDiscovery();
 
     // Takes the updateTerminalDiscovery path — profile data refreshed,
-    // but no re-scoring, no SourceKennel deletion, no SourceKennel upsert.
+    // no re-scoring, no SourceKennel deletion.
     expect(prisma.sourceKennel.deleteMany).not.toHaveBeenCalled();
-    expect(prisma.sourceKennel.upsert).not.toHaveBeenCalled();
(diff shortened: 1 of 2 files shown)


johnrclem:
@codex @claude review

PR #618, opened 2026-04-11
Allow unauthenticated users to open sign-in modal from feedback button

johnrclem:
## Summary
Updated the feedback dialog to allow unauthenticated users to access the feedback feature by presenting a sign-in modal, rather than hiding the feedback button entirely for signed-out users.

## Key Changes
- Imported `SignInButton` from `@clerk/nextjs` to enable authentication flow
- Extracted the button styling into a reusable `triggerClass` constant to maintain consistency across both authenticated and unauthenticated states
- Modified the unauthenticated user handling to render a `SignInButton` wrapper around the feedback button instead of returning `null`
- Updated the authenticated user's dialog trigger to use the extracted `triggerClass` constant

## Implementation Details
- The feedback button now appears for all users, but unauthenticated users will see a sign-in modal when clicked
- The button styling remains consistent between both states using the shared `triggerClass` constant
- Sign-in uses modal mode to keep users on the current page during authentication

https://claude.ai/code/session_01XMt8osrUHcWanDbABho39y

### DIFF
--- a/src/components/feedback/FeedbackDialog.tsx
+++ b/src/components/feedback/FeedbackDialog.tsx
@@ -26,8 +26,15 @@ import { Button } from "@/components/ui/button";
 
 const triggerClass = "text-foreground/80 transition-colors hover:text-foreground";
 
+/**
+ * Footer/menu trigger for submitting user feedback.
+ *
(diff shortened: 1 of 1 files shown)


johnrclem:
@claude @codex review

PR #767, opened 2026-04-17
fix(travel): post-MVP prod QA — 4 surgical fixes

johnrclem:
## Summary

Four QA findings from authenticated prod testing at hashtracks.xyz, all correctness-or-UX-consistency, shipped as a single small PR. `/frontend-design:frontend-design` input shaped the chip + badge decisions before code landed.

## Fixes

**Fix 1 — Confidence label vs evidence-count contradiction (P1 trust regression)**
Prod cards showed `Medium confidence · 0 runs in last 12 weeks` — a direct contradiction of the PRD Confidence Model (Medium requires ≥1 recent run). `scoreConfidence` now downgrades `medium → low` whenever `confirmedEventCount === 0`, regardless of kennel activity. Affected projections drop from Likely into Possible where they belong. One new test pins the gate; one existing test needed a seeded evidence event because it was implicitly covering the bug.

**Fix 2 — Saved chip syncs to URL; dropdown killed (P2 state/UX)**
Editing dates on a previously-saved trip left the chip stuck on `Saved · Your trips`, and the dropdown's `Update with current params` silently overwrote the original trip.

Two changes:
- `useEffect(setSavedId(initialSavedId), [initialSavedId])` — standard React state-sync.
- Per design: split-button dropdown killed entirely. Single-toggle `SavedBadge` — emerald-outlined button that unsaves with a `sonner` Undo action (re-saves with current params, new row). The chip is purely a save-state indicator on the results page; trip management lives on `/travel/saved`.

**Fix 3 — RADIUS ADJUSTED badge for snap-clamp (P2 UX consistency)**
`?r=200` was silently snapped to `?r=100` without UI signal. Two distinct badges now:
- `◆ Routing revised` (filled) — broader-region expand-up
- `◇ Radius adjusted` (open) — tier-snap clamp-down

When both fire, ROUTING REVISED wins visually and the meta line renders the full sequence `<s>200 km</s> → <s>100 km</s> → 300 km`. `snapRadiusToTier` + `RADIUS_TIERS` extracted to `limits.ts` so `page.tsx` can snap server-side on first paint.

**Fix 4 — Delete modal disambiguates same-destination trips (P3 polish)**
Two London trips on different dates produced identical delete modals. The `AlertDialog` now renders the destination on its own line above a monospace-uppercase meta line matching the TripSummary meta treatment. Destructive button relabeled `Delete trip`.

## Deferred to separate GitHub issues

- `travel: 503s on _rsc requests` — needs Vercel log investigation, not a code change
[...]

PR #781, opened 2026-04-17
fix(travel): Undo re-saves + suppress revision badges on no-coverage

johnrclem:
## Summary

Two findings from the re-test of PR #767 that surfaced after merge.

### P2 — Undo doesn't re-save (regression of advertised behavior)

QA repro: click `Saved` to unsave → sonner toast `Removed from saved trips · Undo` → click Undo → toast dismisses, but the trip is still archived in the DB. The restore server action never fired.

**Root cause:** the Undo `onClick` wrapped `restoreTravelSearch` in `startMutation(async () => { ... })`. Sonner's action button renders through a portal, so the click event is a native DOM event rather than a React synthetic event. Async state updates inside a `useTransition` callback launched from that boundary didn't schedule reliably — the server action appeared to no-op silently.

**Fix:** make the onClick directly async; call `restoreTravelSearch` + `setSavedId` without the `useTransition` wrapper. The UX-level indirection from `isMutating` didn't apply to the Undo path anyway (`SavedBadge` isn't visible while `savedId` is null during the Undo window).

### P3 — Revision badges fire on no-coverage empty state

QA repro: Ushuaia search (no kennels at 50 km, no kennels at the 150 km broader-fallback either) rendered `◆ Routing revised · 50 km → 150 km` alongside the `We haven't mapped any hashes here — yet` empty state. The badge falsely implies the expansion found something.

**Fix:** gate `radiusSnapped` and `broaderExpanded` derivations on `!noCoverage`. When there's nothing to show, the EmptyStates copy owns the page and any revision signal becomes misleading noise.

## Test plan

- [x] `npm run lint` no new warnings (pre-existing react-hooks warning from 8fec5ce untouched)
- [x] 195 travel tests still passing
- [ ] Preview browser — Undo path:
  - [ ] Save a trip → click Saved → Undo in the toast → chip flips back to Saved, reload confirms DB row restored with original id
- [ ] Preview browser — no-coverage badge:
  - [ ] Ushuaia search → empty state renders without any ◆/◇ revision badge and no strikethrough in the meta line

### DIFF
--- a/src/components/travel/TripSummary.test.ts
+++ b/src/components/travel/TripSummary.test.ts
@@ -1,0 +1,16 @@
+import { describe, it } from "vitest";
+
+// Placeholder for interaction coverage. The undo/restore flow logic lives
+// in TripSummary's handleRemove handler; the server action paths are
+// covered by actions.test.ts (restoreTravelSearch success/error branches).
+// What's missing is the React-rendering side: that clicking Undo fires the
+// server action, updates savedId on success, dismisses the success toast,
+// and shows an error toast on failure. Pending Travel-Mode RTL setup.
+describe("TripSummary", () => {
+  it.todo("restores savedId and captures analytics when Undo succeeds");
+  it.todo("dismisses the success toast and shows error when Undo fails");
+  it.todo("dismisses the success toast and shows error on network failure");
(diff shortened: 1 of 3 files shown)


johnrclem:
@claude review and address any feedback in the PR

PR #739, opened 2026-04-15
feat(travel): Travel Mode MVP — trip planner for hashers

johnrclem:
## Summary

Travel Mode: a full-trip planner for hashers. Enter destination + date range, get confirmed events, likely projections, and cadence-based possibles grouped by distance tier and day. Guest-friendly; Save Trip requires auth.

**Live at:** `/travel` · `/travel/saved`

## What's in the box

**Data layer**
- New models: `ScheduleRule`, `TravelSearch`, `TravelDestination` (Prisma migration included)
- `ScheduleRule` backfill script populates cadence data for 390+ kennels
- Projection engine: confirmed / high-confidence likely / medium / low cadence-based
- Search service: distance tiering (nearby/area/drive), timezone-aware, dedup

**Server actions (auth + ownership scoped)**
- `saveTravelSearch` — idempotent on coord identity
- `updateTravelSearch` — mutates in-place
- `deleteTravelSearch` — archive
- `findExistingSavedSearch` — SSR recognition of saved trips
- `viewTravelSearch` · `listSavedSearches` · `getDestinationKennelCount` · `resolveDestinationTimezone` · `geocodeDestination`

**Frontend**
- Boarding-pass search form with gate-stamp REQUIRED validation + IATA-style REF codes
- Hero landing with popular destinations + \"near me\" shortcut + Places autocomplete w/ geocoding fallback
- TripSummary: \"You'll catch N confirmed trails\" + Save/Share/Export Calendar
- Result cards grouped by distance tier → day: ConfirmedCard (hareline-parity with title/tz-aware time/weather), LikelyCard with evidence timeline + confidence meter, PossibleSection (collapsed) + PossibleRow (inline)
- Inline filter chips: day-of-week with tooltips, Include Possible toggle
- Saved Trips dashboard at `/travel/saved` with SavedTripCard
- SavedState split button: outlined emerald badge + dropdown (Update / Remove)
- 14 analytics events instrumented

**Nav entry**
- Desktop: Travel between Hareline and Logbook
- Footer: Travel Mode in explore links
- Mobile: Travel promoted to primary bottom nav; Kennels moved to More sheet

**Polish + correctness**
- Dense-result animation-delay clamp (max 1500ms stagger)
- Save-intent sessionStorage handshake (guest → sign-in → auto-save)
- Deterministic ICS UIDs (FNV-1a hash) — fixes multi-event export UID collisions
- Empty states, skeleton loading, timezone-aware trip status

## Stats

71 files changed · +12,429 / -404

[...]

PR #787, opened 2026-04-17
fix(pipeline): canonical Event row per (kennelId, date) — kills Travel/Hareline drift

johnrclem:
## Summary

Phase 2 of the Travel roadmap. Prod audit flagged Travel + Hareline rendering different times/locations for the same real-world run (Easter Hash: 12:30 PM vs 8:00 AM; Marathon: Hangover Trail: 2:30 PM vs 2:00 PM). Root cause: when two sources disagree on startTime/title/sourceUrl, `upsertCanonicalEvent`'s disambiguation falls through and creates a second `Event` row. Both persist; Travel's + Hareline's queries pick different rows when ordering ties.

Fix without squashing the audit trail.

## What changed

- **Schema**: new `Event.isCanonical Boolean @default(true)` + compound index `[kennelId, date, isCanonical]`. Migration `20260417180540_event_is_canonical_flag`.
- **Pure selector**: `pickCanonicalEventId` + `completenessScore` exported from `src/pipeline/merge.ts`. Orders candidates by `trustLevel DESC, completeness DESC, createdAt ASC`. Trust wins categorically over completeness.
- **Pipeline hook**: `recomputeCanonical` runs at the end of every successful `upsertCanonicalEvent`, using `sameDayEvents` the caller already fetched + the just-created row. No extra DB round-trip in the common case; skip-if-unchanged check avoids redundant transactions on chronic-dup kennels.
- **Display filters**: Hareline list + Travel's confirmed/evidence queries now filter on `isCanonical: true` via shared `DISPLAY_EVENT_WHERE` / `CANONICAL_EVENT_WHERE` constants in `src/lib/event-filters.ts`. Individual event detail pages lookup by id — no filter needed.
- **Backfill**: `scripts/dedup-event-rows.ts`. Scans every `(kennelId, date)` with >1 row, runs the selector, flips losers. Dry-run by default; `--apply` to commit. Safe to re-run.

## Tests

- 16 new unit tests cover `completenessScore` + `pickCanonicalEventId` (empty, single, tied trust, tied completeness, tied all, three-way, trust-over-completeness precedence).
- Existing merge tests all pass unchanged — `$transaction` + `updateMany` added to the Prisma mock plus a `[]` default for `mockEventFindMany`.
- 4751 total tests green; TSC clean; lint clean.

## Verification plan

- [x] Unit tests cover the selector
- [x] `/simplify` pass applied (commit a34a0ec)
- [ ] `/codex:adversarial-review` — running in background
- [ ] Preview browser:
  - [ ] Easter Hash renders one consistent time/location across Travel + Hareline
  - [ ] Marathon: Hangover Trail same check
  - [ ] Any destination with known multi-source kennels (e.g. BFM, Boston multi-calendar) doesn't regress
- [ ] Post-merge, run `tsx scripts/dedup-event-rows.ts --apply` against prod DB once to backfill `isCanonical` on historical dup rows

## Affects

This is Hareline correctness debt AND Travel correctness debt — the drift has been visible in both surfaces. Not just a Travel follow-up.

### DIFF
--- a/prisma/migrations/20260417180540_event_is_canonical_flag/migration.sql
+++ b/prisma/migrations/20260417180540_event_is_canonical_flag/migration.sql
@@ -1,0 +1,11 @@
+-- AlterTable
+ALTER TABLE "Event" ADD COLUMN     "isCanonical" BOOLEAN NOT NULL DEFAULT true;
+
+-- CreateIndex
+CREATE INDEX "Event_kennelId_date_isCanonical_idx" ON "Event"("kennelId", "date", "isCanonical");
+
+-- RenameIndex
+ALTER INDEX "ScheduleRule_kennel_rrule_source_key" RENAME TO "ScheduleRule_kennelId_rrule_source_key";
+
+-- RenameIndex
+ALTER INDEX "TravelSearch_id_userId" RENAME TO "TravelSearch_id_userId_key";
--- a/prisma/migrations/20260417184354_drop_redundant_event_kennel_date_index/migration.sql
+++ b/prisma/migrations/20260417184354_drop_redundant_event_kennel_date_index/migration.sql
@@ -1,0 +1,2 @@
+-- DropIndex
+DROP INDEX "Event_kennelId_date_idx";
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -345,8 +345,9 @@ model Event {
   updatedAt         DateTime           @updatedAt
 
   @@index([date])
-  @@index([kennelId, date])
   @@index([parentEventId])
+  // Covers the former `[kennelId, date]` index as a prefix; prefix lookups
+  // still hit it without needing a separate entry.
   @@index([kennelId, date, isCanonical])
 }
 
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -324,13 +324,13 @@ model Event {
   status            EventStatus @default(CONFIRMED)
   isManualEntry     Boolean     @default(false)
   submittedByUserId String?
-  /// When two sources disagree on a (kennelId, date) and the merge pipeline
-  /// can't collapse them into one row, every row is kept for audit fidelity
-  /// but only ONE is canonical — display paths (Travel, Hareline) filter on
+  submittedBy       User?       @relation("SubmittedEvents", fields: [submittedByUserId], references: [id])
+  /// When two sources disagree on (kennelId, date) and the merge pipeline
+  /// can't collapse them into one row, every row persists for audit
+  /// fidelity but only ONE is canonical — display paths filter on
   /// isCanonical: true. Selector: highest trustLevel, then highest
   /// completeness score, then oldest createdAt (stable).
   isCanonical       Boolean     @default(true)
-  submittedBy       User?       @relation("SubmittedEvents", fields: [submittedByUserId], references: [id])
 
   kennel            Kennel             @relation(fields: [kennelId], references: [id])
   parentEvent       Event?             @relation("EventSeries", fields: [parentEventId], references: [id])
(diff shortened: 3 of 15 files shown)


johnrclem:
@claude @codex review

johnrclem:
## Codex adversarial review addressed (commit b19f140)

Ran `/codex:adversarial-review --base 4f78469 --background` after the initial push. Codex flagged three items; two shipped in this follow-up commit, one deferred with rationale.

**Finding 1 (shipped) — `recomputeCanonical` scored pre-update state.** The UPDATE branch enriched fields on the target row, but `recomputeCanonical` then scored `sameDayEvents` (stale). For equal-trust siblings, the sibling's pre-update completeness could win even though the updated row would beat it post-update. → Both update paths now capture the `prisma.event.update()` return and splice the fresh row back into `sameDayEvents` before the canonical recompute. New regression test in `merge.test.ts` covers the equal-trust completeness-shift case.

**Finding 2 (shipped) — isCanonical filter incomplete on public paths.** Travel + Hareline weren't the only user-facing Event queries. → Filter added to:
- `src/app/logbook/actions.ts` — via `commonFilters` so both logbook branches inherit
- `src/app/sitemap.ts` — SEO sitemap emitting dup URLs to search engines would be bad

Admin / merge / `findUnique`-by-id paths deliberately keep no filter: admin needs to see dups to manage them, merge needs all rows to pick canonical, id-based lookups don't create drift.

**Finding 3 (deferred) — DB-level partial unique on `(kennelId, date) where isCanonical=true`.** Non-trivial because `batchMatchedEvents` + double-header logic already produces intentional same-day rows that would collide. Revisit if app-layer drift leaks through.

Updated stats: 806 tests passing (+1 regression).

johnrclem:
## PR review round 2 addressed (commit c6ec7bf)

Must-fix items from the aggregated review (Claude Bot BLOCKER + Codex double-header concern + CodeRabbit outside-diff):

### Double-header regression fixed
Pre-fix, `recomputeCanonical` picked one winner per `(kennelId, date)` and demoted everything else. `upsertCanonicalEvent` intentionally creates a second row when startTime / sourceUrl / title differ (the batchMatchedEvents guard's morning-vs-evening-trail path), so a real double-header would go invisible under `isCanonical: true` filtering.

Introduced **signature grouping** inside the selector — `pickCanonicalEventIds` now returns `Set<string>`, one canonical per `(startTime, sourceUrl, title)` group. Dup-drift collapses, genuine double-headers survive. Three new regression tests cover preserving double-headers, collapsing dup-drift, and mixed same-day slots.

### Kennel display paths
Added `isCanonical: true` to the four remaining public paths Claude Bot flagged:
- `src/app/kennels/[slug]/page.tsx` (BLOCKER — full event list)
- `src/app/page.tsx` (home upcoming-runs widget)
- `src/app/kennels/page.tsx` (directory next-event)
- `src/app/kennels/region/[slug]/page.tsx` (region map)

### Polish items
- Coord pair counts as one completeness unit (half a pair is useless).
- Dropped redundant `@@index([kennelId, date])` — covered by the 3-col index as a prefix. New migration `20260417184354_drop_redundant_event_kennel_date_index`.
- `recomputeCanonical` filters writes to rows whose flag actually flips.
- Backfill script summary log uses "were" instead of bare empty string.

### Still deferred (documented)
- DB-level partial unique on `(kennelId, date) where isCanonical=true` — blocked by intentional double-header rows, revisit if drift leaks through.

### Deploy runbook
Post-merge: one `tsx scripts/dedup-event-rows.ts --apply` against prod to backfill `isCanonical` on historical dup rows. Same pattern as PR #739's ScheduleRule backfill.

**4755 tests passing**, TSC clean, lint clean.

Not addressing (deferred):
- Batched `findMany` in the dedup script — fine at prod scale (~200 groups × 30ms = 6s one-shot).

Issue #817, opened 2026-04-20
HARRIER_CENTRAL adapter attributes all kennel events to "Singapore Sunday H3 Harrier Central" source regardless of kennel


johnrclem:
## Summary

All events ingested via the `HARRIER_CENTRAL` adapter are being attributed in the admin panel to the "Singapore Sunday H3 Harrier Central" source record, even for completely different kennels (Tokyo H3, Morgantown H3). This means the `sourceId` foreign key on event records created by the Harrier Central adapter always resolves to the SH3-SG source rather than the kennel-specific source.

## Affected Kennels / Events

**Tokyo H3** (12 events, all attributed to "Singapore Sunday H3 Harrier Central"):
- Example: https://www.hashtracks.xyz/kennels/tokyo-h3
- Admin events filter: `kennelId=cmndkck5u0004t6hnmnj72iki`

**Morgantown H3** (runs #133–137 all show "Singapore Sunday H3 Harrier Central" as source):
- Admin events filter: `kennelId=cmnouyosr0006jghnm34olk6y`

## Source URL

All three HARRIER_CENTRAL sources share the same base URL: `https://harriercentral...`

- "Singapore Sunday H3 Harrier Central": config: `kennelUniqueShortName: "SH3-SG"`
- "Tokyo H3 Harrier Central": config: `cityNames: "Tokyo"`
- "Morgantown H3 Harrier Central": config: `kennelUniqueShortName: "MH3-WV"`

## Suspected Adapter

`HARRIER_CENTRAL` (multi-kennel aggregator adapter)

## Field(s) Affected

`sourceId` on event records (wrong source linked)

## Current Behavior

In the admin panel, "Source(s)" column shows:
- `Singapore Sunday H3 Harrier Central` for Tokyo H3 events
- `Singapore Sunday H3 Harrier Central` for Morgantown H3 events (in addition to their Google Calendar source)

## Expected Behavior

- Tokyo H3 events should show "Tokyo H3 Harrier Central" as source
- Morgantown H3 events should show "Morgantown H3 Harrier Central" as source

## Fix Hypothesis

The HARRIER_CENTRAL adapter likely looks up the source record by matching on the shared `url` field (all three sources have the same base API URL). Since SH3-SG was created first or sorts first, it always wins the lookup. The adapter needs to disambiguate by matching on both `url` + `config.kennelUniqueShortName` (or `config.cityNames`), or by storing a unique sourceId per kennel config entry rather than per API URL.

[...]

Issue #789, opened 2026-04-17
[Boston H3] Shared bostonhash@gmail.com calendar cross-attributes sibling Boston-area kennels to Boston H3


johnrclem:
### Summary

The `bostonhash@gmail.com` GOOGLE_CALENDAR is a shared calendar serving the entire Boston hash community, not just Boston H3. Per https://www.bostonareahashes.org the area has at least these active kennels:

- Boston H3 ← this is who HashTracks attributes everything to
- Boston Moon H3 (BMH3)
- Boston Ball Buster Hardcore H3
- Pink Taco Trotters
- Eager4Beaver
- ZigZag
- Beantown City Hash

And the shared calendar visible at https://www.bostonhash.com/calendar contains events for several of these sibling kennels. Two failure modes result:

### Failure mode A: sibling-kennel events attributed to Boston H3

Example: "Boston Moom" on Fri, Mar 20 2026 (see https://www.hashtracks.xyz/kennels/boston-h3?tab=past) is filed under Boston H3 but is actually a Boston Moon H3 event. A Facebook comment from Boston H3 themselves confirms "Boston Moon H3" is a separate kennel ("Not sure when Boston Moon H3 will be doing trail...").

Other likely cross-attributed titles on the same calendar:
- "Taco Marathon Pre-Pre-Pre-Prelube" (Apr 14, 2026) — Pink Taco Trotters
- "Moon Marathon Pre-Pre Lube" (Apr 16, 2026) — Boston Moon H3

### Failure mode B: sibling-kennel events dropped entirely

"Beantown #276" (Apr 1), "Beantown #277" (Apr 8), "Beantown #278" (Apr 15) all appear on the calendar but `https://www.hashtracks.xyz/kennels/beantown-h3` returns 404 and these events don't appear under Boston H3 either, so they're being dropped by the adapter. This is actually the desirable behavior if Beantown ever gets its own kennel record, but right now Beantown hashers browsing HashTracks can't find their kennel.

### Suspected Adapter
`GOOGLE_CALENDAR` source `bostonhash@gmail.com`.

### Fix Hypothesis
The adapter is treating the whole calendar as Boston H3's and appears to use a soft filter (prefix match? title regex?) that lets some non-BH3 titles slip through ("Boston Moom", "Taco Marathon") while filtering others out ("Beantown #"). Options:

[...]

PR #822, opened 2026-04-20
fix(seed): source identity is (name,type), not (url,type) (#817, #789)

johnrclem:
Two source-attribution bugs shared a root cause: records that legitimately
share a URL (config-driven REST APIs, shared calendars) were being
collapsed or cross-attributed because seed and adapter logic keyed off
the wrong identifier.

#817 — The 3 HARRIER_CENTRAL sources (Tokyo H3, Morgantown H3, Singapore
Sunday H3) share the base API URL https://harriercentralpublicapi.../;
prisma/seed.ts matched existing records by `(url, type)` first, so each
subsequent seed run UPDATED the prior row's name + config instead of
inserting a new one. Prod collapsed to a single row named "Singapore
Sunday H3 Harrier Central" with all 3 kennels linked to it. Narrow the
identity check to `(name, type)` only — url is still synced via the
existing scalar-field loop further down. Adds a seed-data invariant test
to guard against re-collapse.

#789 — The bostonhash@gmail.com Google Calendar feeds 5+ Boston-area
kennels (Boston H3, Boston Moon, Beantown, Pink Taco Trotters, Boston
Ball Buster), but pattern matching lived hardcoded in
google-calendar/adapter.ts and didn't cover observed title variants like
"Boston Moom" typo or "Taco Marathon" (no "Pink" prefix). Migrate those
patterns into source config (matching Philly, Chicago, BFM, Princeton,
etc.) and extend them to cover the failing cases. Adds adapter tests for
each failure mode from the issue body.

Also adds scripts/split-harrier-central-sources.ts — a dry-run-by-default
one-shot script that splits the collapsed prod HC row into 3 distinct
rows and re-parents SourceKennel links. Must be run manually against
prod before the next HC scrape cycle so Tokyo/Morgantown events attribute
correctly.

https://claude.ai/code/session_01V7m1qsjsrqvy5GhTXBkSjG

### DIFF
--- a/prisma/seed-data/sources.test.ts
+++ b/prisma/seed-data/sources.test.ts
@@ -1,4 +1,3 @@
-import { describe, it, expect } from "vitest";
 import { SOURCES } from "./sources";
 
 describe("SOURCES seed data invariants (#817 regression guard)", () => {
--- a/prisma/seed-data/sources.test.ts
+++ b/prisma/seed-data/sources.test.ts
@@ -1,0 +1,41 @@
+import { describe, it, expect } from "vitest";
+import { SOURCES } from "./sources";
+
+describe("SOURCES seed data invariants (#817 regression guard)", () => {
+  it("every (name, type) pair is unique", () => {
+    // prisma/seed.ts:363 looks up existing sources by (name, type) as the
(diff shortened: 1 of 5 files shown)


johnrclem:
@codex @claude review

PR #823, opened 2026-04-20
fix(db): enforce (name, type) as Source identity at the DB layer

johnrclem:
## Summary

Follow-up to #822. That PR fixed the [#817](https://github.com/johnrclem/hashtracks-web/issues/817) HARRIER_CENTRAL source collapse at the seed-upsert layer by keying on `(name, type)` instead of `(url, type)`. This PR closes the same gap at the DB layer so admin-created sources, manual SQL, or future seed edits can't silently reintroduce the duplicate.

- Add `@@unique([name, type])` on the `Source` model.
- Migration is a single `CREATE UNIQUE INDEX`; ran a prod audit before authoring and confirmed zero existing `(name, type)` duplicates on Railway.
- Translate `P2002` on the new constraint into a friendly validation message in the admin write paths (`createSource`, `updateSource`, `approveProposal`) so users see _"A source with that name and type already exists."_ instead of a raw Prisma exception.

## Why this + the seed fix

The seed-side fix in #822 is sufficient to prevent _seed_ runs from collapsing sources — but it leaves admin-created sources unprotected. If an admin inadvertently types an existing `(name, type)` pair, the DB will accept it today. With this index, the DB rejects it and the admin UI surfaces a clean error.

## Rollout risk

`migrate deploy` runs the `CREATE UNIQUE INDEX` in a transaction. If an admin creates a colliding source between the prod audit and the index build, the deploy transaction will fail cleanly and roll back. Risk window is seconds; if it triggers we retry after resolving the collision. `CONCURRENTLY` can't be used inside a migration transaction.

## Verification

- Prod `(name, type)` audit: 0 duplicates
- Local DB audit: 0 duplicates
- Migration applies on local DB + a fresh shadow DB (full `prisma/migrations/` replay)
- `npx tsc --noEmit` clean
- `npm run lint` — 0 errors (pre-existing warnings unchanged)
- `npm test` — 211 files, 4894 tests passing
- Codex adversarial review — 2 findings, both addressed (race window documented above; P2002 handling added to admin paths)

## Test plan

- [ ] CI green (type check + lint + tests)
- [ ] Preview deploy succeeds (Vercel runs `prisma migrate deploy`)
- [ ] Admin UX: create a source with a duplicate `(name, type)` — should show friendly error, not Prisma internals

### DIFF
--- a/prisma/migrations/20260420203030_source_name_type_unique/migration.sql
+++ b/prisma/migrations/20260420203030_source_name_type_unique/migration.sql
@@ -1,0 +1,6 @@
+-- Enforce (name, type) as Source identity at the DB layer.
+-- prisma/seed.ts::ensureSources already upserts by (name, type) in app code;
+-- this index makes the invariant impossible to violate via admin creation,
+-- manual SQL, or future seed edits that collide on (name, type).
+-- See #817 for the HARRIER_CENTRAL collapse that motivated this guard.
+CREATE UNIQUE INDEX "Source_name_type_key" ON "Source"("name", "type");
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -205,6 +205,10 @@ model Source {
   eventLinks    EventLink[]
   createdAt     DateTime       @default(now())
   updatedAt     DateTime       @updatedAt
+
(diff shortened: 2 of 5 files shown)


johnrclem:
@codex @claude review

Issue #864, opened 2026-04-21
reconcile: normalize scraped/candidate slot-key date format with shared helper


johnrclem:
## Context

Follow-up from PR #863 review (CodeRabbit inline comment).

In \`src/pipeline/reconcile.ts\`, the slot key is built two different ways:

- **Scraped side** (line ~55): \`\${kennelId}:\${event.date}\` — uses the raw \`RawEventData.date\` string from the adapter.
- **DB side** (line ~139): \`\${event.kennelId}:\${dateStr}\` where \`dateStr = event.date.toISOString().split(\"T\")[0]\`.

Today these always agree because adapters emit \`event.date\` as \`\"YYYY-MM-DD\"\` and DB \`Event.date\` is stored as UTC noon, so \`toISOString().split(\"T\")[0]\` also produces \`\"YYYY-MM-DD\"\`. But the two sides are textually different constructions and nothing enforces the invariant.

This is a **pre-existing** pattern (predates #863); the sourceUrl-drift fix left it untouched. Flagging for separate hardening.

## Proposed fix

Introduce a shared helper in \`src/pipeline/reconcile.ts\` (or \`src/adapters/utils.ts\`):

\`\`\`ts
function toSlotDate(date: string | Date): string {
  if (date instanceof Date) return date.toISOString().split(\"T\")[0];
  if (!/^\d{4}-\d{2}-\d{2}$/.test(date)) {
    throw new Error(\`Invalid slot date format: \${date}\`);
  }
  return date;
}
\`\`\`

Use it on both sides of the slot-key build. The runtime validation also catches future adapter regressions (e.g. someone emitting an ISO timestamp or \`\"2026-4-1\"\`).

## Acceptance

- [ ] Both slot-key builds go through \`toSlotDate\`
- [ ] Test: scraped row with malformed \`event.date\` throws or is skipped (not silently mismatched)
- [ ] Test: existing reconcile behavior unchanged for well-formed inputs

## Non-goals

- No changes to the merge pipeline key construction — scope is reconcile only.
- No schema changes.

johnrclem:
Merged via PR #867 (commit b7da518defc128420143533bc7fb3d79f550dc91).

PR #867, opened 2026-04-22
fix(reconcile): normalize slot-key dates through shared helper (GH #864)

johnrclem:
## Summary

- Follow-up hardening to PR #863. Normalizes scraped-side and DB-side slot-key date construction in `src/pipeline/reconcile.ts` through a new `toIsoDateString` helper that matches merge's `parseUtcNoonDate` semantics exactly — so every date shape merge accepts (loose forms like `"2026-4-1"`, offset timestamps like `"…T23:30:00-05:00"`, space-separated times, overflow dates like `"2026-02-31"` → `"2026-03-03"`) produces the key merge would write.
- Fail-safe for unparseable dates: exclude the whole kennel from cancellation this run instead of dropping the row (which would orphan the canonical → false cancellation, strictly worse than the original mismatch). Surface via new `kennelsSuppressedForBadDate` field on `ReconcileResult`, persisted in scrape log diagnostic context, and `console.warn`'d at scrape level.
- 16 new tests (11 helper + 5 reconcile regression). No data cleanup needed — prophylactic hardening.

## Why this matters

`RawEventData.date` is documented as `"YYYY-MM-DD"` but has no runtime enforcement. If an adapter silently drifts to an ISO timestamp (WordPress API returns `"2026-03-29T15:00:00"`), offset form, or loose numeric (`"2026-4-1"`), the scraped-side key no longer matches the DB-side key, reconcile sees the canonical as orphaned, and flips it to CANCELLED — the exact false-cancellation class #863 just fixed, but without the URL-drift smoking gun.

Two rounds of `/codex:adversarial-review` caught an important correctness issue mid-review: an earlier strict-regex cut of the helper would have silently disabled stale-event cleanup for any kennel whose adapter emitted a merge-compatible-but-strict-regex-incompatible form. The final design delegates to `parseUtcNoonDate` so reconcile and merge can't drift.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors
- [x] `npm run lint` — 0 errors (12 pre-existing unrelated warnings)
- [x] `npm test` — 4972 pass (added 16 new tests)
- [x] Two rounds of `/codex:adversarial-review` — high finding resolved, medium (promoting suppression to a first-class health alert) filed as follow-up since it requires a new `AlertType` enum + migration + admin UI surface beyond this PR's scope
- [x] `/simplify` pass — cleaned up hot-path regex, hoisted `suppressedKennels` array, tightened docstring

## Follow-up

Related task filed: promote `kennelsSuppressedForBadDate` to a first-class health alert so persistent adapter drift generates a visible alert instead of only a log warning + diagnostic context entry.

Closes GH #864.

### DIFF
--- a/src/lib/date.test.ts
+++ b/src/lib/date.test.ts
@@ -1,4 +1,4 @@
-import { getTodayUtcNoon, parseUtcNoonDate } from "./date";
+import { getTodayUtcNoon, parseUtcNoonDate, toIsoDateString } from "./date";
 
 describe("getTodayUtcNoon", () => {
   it("returns a number (milliseconds timestamp)", () => {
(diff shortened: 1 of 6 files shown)


johnrclem:
@claude review

PR #865, opened 2026-04-22
fix(pipeline): dedup cross-source canonical events on (runNumber, sourceUrl)

johnrclem:
## Summary

Fixes #826 — two SFH3 sources (ICAL_FEED + HTML_SCRAPER) were producing two canonical Event rows for GPH3 run #1704 on 2026-04-23, both with `isCanonical=true` (titles "GPH3 Run #1704" at trust=8 and "Almost Old Enough to be a Museum Piece" at trust=7).

Root cause: the upsert matcher keyed on `sourceUrl` alone and the canonical selector keyed on a signature that couldn't collapse differently-titled rows at the same runNumber. QStash fan-out made the race visible — two sources scraping simultaneously each CREATE'd a row.

## Changes

**`src/pipeline/merge.ts`**
- `eventSignature()` now prefers `(runNumber, sourceUrl)` when runNumber is present, falling back to `(startTime, sourceUrl, title)`. Collapses cross-source rows for the same run.
- `pickCanonicalEventIds()` is status-aware: CANCELLED rows are demoted out of the selection pool when any live row exists, so a stale CANCELLED row at trust=8 won't beat a live CONFIRMED row at trust=7.
- `upsertCanonicalEvent()` adds a runNumber match step after URL match, gated on `runMatches.length === 1` so multi-part events at distinct URLs don't collide. Deterministic `orderBy: [createdAt asc, id asc]` keeps matcher behavior stable across retries.
- Lower-trust CANCELLED restore now mirrors `status` in-memory so `recomputeCanonical` sees the just-restored row as live.

**`scripts/dedup-event-rows.ts`**
- Select `status` and annotate `[MIXED-STATUS]` groups in log output for ops visibility.

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] `npm run lint` — 0 errors (12 pre-existing warnings)
- [x] `npm test` — 4951 tests pass (221 in merge.test.ts, up from 217 baseline)
- [x] New merge tests cover: `(runNumber, sourceUrl)` signature matrix, status-aware selector (CONFIRMED wins over higher-trust CANCELLED), URL-less incoming with unique runNumber matches the row, ambiguous runNumber (multi-part) skips the runNumber step, ambiguous runNumber with no discriminator creates a new row, lower-trust restore mirrors status in-memory
- [x] Ran `dedup-event-rows.ts --apply` on prod pre-compaction — 3 slots flipped, all verified correct via Railway psql
[...]

Issue #826, opened 2026-04-21
GPH3 — Run #1704 ingested twice (duplicate event): ICAL_FEED and HTML_SCRAPER both produce separate records from same source URL


johnrclem:
## Finding

GPH3 run #1704 appears twice on the hareline on April 23, 2026, as two distinct HashTracks events with different titles but the same source URL, same run number, same hare, same date, and same location.

**Duplicate Event A (ICAL_FEED):** https://www.hashtracks.xyz/hareline/cmo64sfag000804jvwakcfawg
   title: "GPH3 Run #1704"
   description: "Hare: Blow Queen"

**Duplicate Event B (HTML_SCRAPER):** https://www.hashtracks.xyz/hareline/cmo64sfbq000304jm9m1r5d1n
   title: "Almost Old Enough to be a Museum Piece"
   description: null

**Both have:**
- sourceUrl: `https://www.sfh3.com/runs/6516`
- runNumber: 1704
- haresText: "Blow Queen"
- startTime: "18:15"
- locationName: "Parking lot for the Randall Museum, San Francisco, CA"

**Suspected Adapters:** `ICAL_FEED` (SFH3 MultiHash iCal Feed) + `HTML_SCRAPER` (SFH3 MultiHash HTML Hareline)

**Source URL confirmed:** https://www.sfh3.com/runs/6516

## Root Cause Hypothesis

The sfh3.com aggregator site is the data source for both a Bay Area iCal feed (`sfh3.com/calendar.ics?kennels=all`) and a Bay Area HTML hareline scraper (`sfh3.com/runs?kennels=all`). Both ingest the same GPH3 event independently, and the deduplication logic doesn't catch the collision because:

1. The iCal event uses the SFH3-generated title format "GPH3 Run #1704"
2. The HTML scraper finds the actual event title "Almost Old Enough to be a Museum Piece"
3. The deduper may not match on runNumber+kennelId+date when titles differ

**Fix Suggestion:** Add a deduplication check on `(kennelId, runNumber, date)` or `(kennelId, sourceUrl)`, keeping the higher-trustLevel record (HTML_SCRAPER trustLevel 7) and merging its data into the surviving event.

---
*Filed by automated chrome hareline audit — 2026-04-21*

Issue #786, opened 2026-04-17
[Brooklyn H3] Kennel profile missing Instagram link


johnrclem:
### Summary

Brooklyn H3's kennel profile on HashTracks shows a Facebook link (correctly pointing at the Brooklyn-specific group, `facebook.com/groups/brooklynh3/`) but no Instagram link. The NYCH3 umbrella that hosts Brooklyn H3's source page surfaces a shared Instagram handle that also covers Brooklyn H3 content.

### Fields Affected
- Kennel Instagram URL

### Expected Value (verbatim from source)
```
https://www.instagram.com/nychashhouse/
```

Linked from the footer/sidebar of https://hashnyc.com/area-hashes/brooklyn-h3/ (`Instagram :: https://www.instagram.com/nychashhouse/`). Same caveat as the logo issue: this handle is shared across the six NYC kennels, so if we attach it to Brooklyn H3 we'll also want to attach it to NYCH3, NAHHH, GGFM, Queens H3, and LIL for consistency.

### Fix Hypothesis
The hashnyc.com HTML_SCRAPER is extracting events but not the umbrella-level social links. Either (a) extend the adapter to grab the footer `Instagram` link when enriching any kennel whose source page lives under hashnyc.com, or (b) set `Kennel.instagramUrl` manually on all six NYCH3-family kennels. A Kennel.socialKind = FAMILY flag (see logo issue) would let the UI label shared handles transparently.

PR #879, opened 2026-04-22
fix(merge): collapse cross-source URL-less/URL-bearing runNumber races

### LINKED ISSUE #866: Cross-source canonical dedup: race path when one source has no sourceUrl
## Problem

Follow-up from #826 (fixed in PR #865). The new \`eventSignature()\` keys on \`(runNumber, sourceUrl)\` when runNumber is present:

\`\`\`ts
if (e.runNumber != null) return \`run#\${e.runNumber}::\${url}\`;
\`\`\`

If two sources race (QStash fan-out) for the same kennel+date and one emits \`runNumber + no URL\` while the other emits \`runNumber + URL\`, the two rows produce **different** signatures:

- \`run#1704::\` (URL-less iCal)
- \`run#1704::https://www.sfh3.com/runs/6516\` (HTML_SCRAPER)

\`pickCanonicalEventIds()\` sees them as separate signature groups and keeps **both** canonical, reproducing the original #826 symptom for this specific configuration.

## When it triggers

- Source A scrape lands first → CREATE (URL-less or with URL)
- Source B scrape starts before Source A's \`recomputeCanonical\` completes (both find 0 sameDayEvents)
- Source B CREATEs a second row
- Neither row's \`sourceUrl\` matches the other's → upsert matcher misses
- Different signatures → selector keeps both canonical

The sequential case is handled correctly by the \`upsertCanonicalEvent\` runNumber matcher step (added in #865), but the true concurrent-create race path is not.

## Pre-existing behavior

This is **not** introduced by #865 — the old \`time::url::title\` signature also kept mismatched-URL rows split. But it's a known gap worth addressing before an ICAL+HTML pair on another kennel triggers it.

## Possible fixes

1. **URL-agnostic signature when runNumber is present** — \`run#\${runNumber}\` alone. Risk: collapses genuine multi-part events (AVLH3 #786 Part B + Part C at distinct URLs). Would need a secondary discriminator (startTime?) to preserve multi-part.
2. **Post-create reconcile pass** — after each upsert, scan \`sameDayEvents\` for another row with the same \`(kennelId, date, runNumber)\` and distinct URL; merge if no secondary discriminator distinguishes them.
3. **Lock by \`(kennelId, date)\` during scrape** — prevents the race entirely, at cost of concurrency.

## Repro

Simulate the race by kicking off two scrapes (different sources, same kennel, same run #) with QStash, one emitting a URL-less ICAL row and one emitting an HTML_SCRAPER row with URL. Verify two canonical rows in \`Event\` table for the same (kennelId, date).

## References

- PR #865 discussion: https://github.com/johnrclem/hashtracks-web/pull/865 (claude-review and my reply both flagged this)
- Original collision: #826
- Relevant code: \`src/pipeline/merge.ts\` \`eventSignature\` and \`pickCanonicalEventIds\`

johnrclem:
## Summary

Closes #866.

- Adds `collapseRunNumberRaces()` in `src/pipeline/merge.ts`: after signature grouping and before canonical selection, absorb a URL-less signature group into its URL-bearing peer when they share a runNumber and startTimes don't conflict.
- Preserves multi-part events (AVLH3 #786 Part B/C) — collapse only fires when one side is URL-less; two URL-bearing groups stay split.
- Conservative fallback: ambiguous topologies (URL-less + 2+ compatible URL-bearing peers) leave all groups canonical.

## Why

Two scrape sources fan out concurrently via QStash. If one emits `sourceUrl` populated and the other doesn't, `eventSignature()` produces `run#N::` vs `run#N::<url>` and both survive — the exact #826 symptom this issue is following up on. The PR #865 upsert matcher only handles the sequential case; this closes the concurrent-create race by letting the post-group reconcile pass in `pickCanonicalEventIds` self-heal on the next scrape.

## Implementation notes

- Compatibility is checked over every row's startTime in each group, not just the representative — signature ignores startTime when runNumber is set, so one URL-less group can contain rows with different times. Required to keep behavior stable against input order.
- Chose option 2 from the issue (post-create reconcile pass) over option 1 (URL-agnostic signature) because option 1 risks collapsing genuine multi-part events.
- Pure logic change in a pure function — no schema or migration needed.

## Test plan

- [x] `npm run lint` clean (0 errors)
- [x] `npm test` — 5009 pass, 238 in `merge.test.ts`
- [x] New regression cases cover: URL-less + URL-bearing collapse (matching + blank startTimes), conflicting startTimes stay split, multi-row URL-less group order-independence (3 permutations), internal time conflict blocks collapse, AVLH3 multi-part preservation, ambiguous three-group topology
- [ ] Post-merge: watch for a drop in same-runNumber canonical dupes surfaced by future reconcile audits

### DIFF
--- a/src/pipeline/merge.test.ts
+++ b/src/pipeline/merge.test.ts
@@ -2148,6 +2148,40 @@ describe("pickCanonicalEventId", () => {
     expect(canonicals.has("url-bearing")).toBe(true);
   });
 
+  it("#866 URL-less internally conflicted times + URL-bearing blank-time peer blocks collapse", () => {
+    // Codex review regression: URL-less group has {09:00, 18:00}, URL-bearing
+    // peer has null startTime. A blank peer cannot disambiguate which of the
+    // two URL-less rows it matches, so collapse must be blocked.
+    const urlLessMorning = candidate({
+      id: "url-less-morning",
+      runNumber: 99,
+      title: "Run #99",
+      startTime: "09:00",
+      sourceUrl: null,
+    });
+    const urlLessEvening = candidate({
+      id: "url-less-evening",
+      runNumber: 99,
+      title: "Run #99",
+      startTime: "18:00",
+      sourceUrl: null,
+    });
+    const urlBearingBlank = candidate({
+      id: "url-bearing-blank",
+      runNumber: 99,
+      title: "Run #99",
+      startTime: null,
+      sourceUrl: "https://example.com/99",
+    });
+    const canonicals = pickCanonicalEventIds([
+      urlLessMorning,
+      urlLessEvening,
+      urlBearingBlank,
+    ]);
+    expect(canonicals.size).toBe(2);
+    expect(canonicals.has("url-bearing-blank")).toBe(true);
+  });
+
   it("#866 ambiguous URL-less + two URL-bearing peers preserves all groups", () => {
     // Conservative fallback: if the URL-less row could belong to either of
     // two URL-bearing multi-part rows, don't guess — keep all three
--- a/src/pipeline/merge.test.ts
+++ b/src/pipeline/merge.test.ts
@@ -2033,11 +2033,152 @@ describe("pickCanonicalEventId", () => {
       rowB: { id: "newer", ...DUP_SIG, trustLevel: 8, status: "CANCELLED" as const, createdAt: new Date("2026-03-01T00:00:00Z") },
       expectedCanonicals: ["newer"],
     },
+    {
+      // #866 cross-source race path: one source emits (runNumber, no URL),
+      // another emits (runNumber, URL). Pre-fix their signatures differ
+      // (`run#N::` vs `run#N::<url>`) and both survive as canonical.
+      name: "#866 URL-less + URL-bearing at same runNumber + compatible startTime collapses",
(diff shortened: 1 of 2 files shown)


johnrclem:
@codex @claude review

johnrclem:
Addressed review feedback in e8a019f8:

**Codex (real bug) — fixed.** `timesCompatible` now only honors the empty-side shortcut when the other side has ≤1 concrete time. A URL-less group with `{09:00, 18:00}` paired with a URL-bearing peer that has `startTime = null` no longer collapses. Added regression test `#866 URL-less internally conflicted times + URL-bearing blank-time peer blocks collapse`.

**Gemini (subset relaxation) — declined.** Kept strict set-equality per Claude-bot's reasoning: `eventSignature` already partitions by `(runNumber, trimmedUrl)`, so a single URL-bearing group rarely accumulates multiple distinct startTimes in the first place. Relaxing to subset would recreate option-1 risk of collapsing genuine multi-part events. Documented the choice inline.

**Claude-bot (in-place mutation) — fixed.** Replaced `groups[peers[0]].push(...urlLess)` with `[...groups[peers[0]], ...urlLess]` so we don't mutate the bucket still referenced by the outer signature map.

**Claude-bot (trust-level footgun) — documented.** Added a code comment noting that if trust levels are ever rebalanced such that a URL-less source outranks its URL-bearing peer, `Event.sourceUrl` would end up null despite a URL-bearing sibling. No source combination produces this today.

Tests: 239 merge tests pass, `tsc --noEmit` clean.

PR #931, opened 2026-04-24
fix(merge): restore CANCELLED events without startTime (#874)

### LINKED ISSUE #874: Dublin H3 — Nash Hash event (Jul 3–5 2026) missing from HashTracks; source hareline has entry not ingested
Dublin H3 (dublinhhh.com) lists a Nash Hash event on their hareline that does not appear in HashTracks.

Source URL: https://dublinhhh.com/hareline

Source hareline shows:
- Date: Friday-Sunday July 3-5 2026
- Hash: Dublin H3 DH3 Nash Hash Dublin
- Location: TBD

HashTracks upcoming events for Dublin H3: Only shows 3 events (Apr 27 #1671, May 4 #413, Sep 14 #???) - the July 3-5 Nash Hash is absent.

Kennel HashTracks page: https://www.hashtracks.xyz/kennels/dublin-h3

Suspected Adapter: HTML_SCRAPER (dublinhhh.com/hareline)
Field(s) Affected: Event not ingested at all

CLI Fix Hypothesis: The Dublin H3 HTML scraper likely fails to parse multi-day events (date range "Friday-Sunday 3-5 July" format) or skips events where the location field contains an event name (DH3 Nash Hash Dublin) rather than a venue. The adapter should handle date-range rows and ingest them with the start date as event date.

johnrclem:
## Summary

Dublin Nash Hash (issue #874) stayed CANCELLED even after the adapter started returning it again. Root cause is in `refreshExistingEvent` in `src/pipeline/merge.ts`: when a processed=true fingerprint matches, the function early-returned if `composeUtcStart()` returned null — which happens whenever the scraped row has no `startTime`. That early return sat *before* the CANCELLED→CONFIRMED restore check, so the restore never ran.

Dublin's hareline renders "3–5 July 2026" with no time, so every scrape hit the early-return path. Once a prior reconcile cycle cancelled the event, no subsequent scrape could bring it back — `upsertCanonicalEvent`'s restore branch is unreachable once the RawEvent is marked processed.

## Fix

- Move `prisma.event.findUnique` above the composedUtc guard.
- Compute `shouldRestore` and `shouldRefreshDateUtc` independently; only skip the update when both are false.
- Only include `dateUtc`/`timezone` in the update payload when we actually have a composed time.

Net behaviour: a scraped row with no startTime now restores a CANCELLED canonical without touching its existing dateUtc/timezone.

## Test plan

- [x] New regression test in `src/pipeline/merge.test.ts` exercises the refreshExistingEvent restore path with `startTime: undefined` — asserts `result.restored === 1` and the update payload is exactly `{ status: "CONFIRMED" }` (no dateUtc clobber).
- [x] All 240 merge tests pass locally.
- [x] `npx tsc --noEmit` + `npm run lint` clean.
- [ ] Post-merge: trigger Dublin Hash scrape from admin UI; confirm the Nash Hash event flips CONFIRMED.

Closes #874

### DIFF
--- a/src/pipeline/merge.test.ts
+++ b/src/pipeline/merge.test.ts
@@ -682,6 +682,36 @@ describe("double-header support", () => {
       }),
     );
   });
+
+  it("restores a CANCELLED row via refreshExistingEvent even when scraped row has no startTime (#874)", async () => {
+    // Dublin Nash Hash regression: the hareline row "3–5 July 2026" has no
+    // time, so composeUtcStart returns null. Without this fix, the early
+    // return in refreshExistingEvent bypassed the CANCELLED→CONFIRMED restore
+    // for processed=true fingerprint matches, leaving the event cancelled
+    // indefinitely once it had been cancelled by a prior reconcile cycle.
+    mockRawEventFind.mockResolvedValueOnce({
+      id: "raw_existing",
+      processed: true,
+      eventId: "evt_cancelled",
+    } as never);
+    vi.mocked(prisma.event.findUnique).mockResolvedValueOnce({
+      trustLevel: 5,
+      dateUtc: null,
+      timezone: "Europe/Dublin",
+      status: "CANCELLED",
+    } as never);
+    mockEventUpdate.mockResolvedValueOnce({} as never);
+
+    const result = await processRawEvents("src_1", [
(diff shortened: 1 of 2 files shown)


johnrclem:
@gemini @claude review

PR #951, opened 2026-04-24
perf(cron): stagger QStash fan-out to eliminate CPU spike from 429 retries

johnrclem:
## Summary

The cron dispatcher was publishing ~120 scrape messages via `Promise.all()` with no QStash `delay`, so all scrapes landed at once, saturated the residential proxy, and triggered HTTP 429 on ~2/3 of requests. The wait-then-retry path burned Vercel Function CPU (~92s in a 5-minute window, flagged by observability as an anomaly).

- Assigns each outbound message a deterministic stagger via `delay: staggerDelaySeconds(i, n)`, spreading delivery across a 240s window (~1 publish / 2 s for 120 sources).
- Stays well inside the existing 10-minute `BUFFER_MS` schedule tolerance, so no source misses its interval.
- Extracts the pure helper + constant to `src/pipeline/schedule.ts` alongside `BUFFER_MS` for co-location and unit testing.
- Orders `prisma.source.findMany` by `id` so the stagger index→source mapping is reproducible across runs.
- `force=true` (operator-initiated backfill / incident response) bypasses the stagger and dispatches immediately.

## Verification

- [x] `npm run lint` — 0 errors
- [x] `npm test` — 5084 tests pass (4 new: stagger math edges, force bypass, N>241 duplicate-delay regime, `STAGGER_WINDOW_SECONDS * 1000 < BUFFER_MS` invariant)

## Post-deploy signals to watch

- [ ] After the next 06:00 UTC cron run, Vercel active-CPU 5-min max drops from ~92s toward prior baseline (~5–10s per bucket × 5 buckets)
- [ ] Residential-proxy 429 rate drops well below 67% (check NAS proxy logs)
- [ ] No increase in scrape failures or stale/reconciled events

## Out of scope (flagged for follow-up)

- `Retry-After` header parsing in [src/adapters/safe-fetch.ts](src/adapters/safe-fetch.ts) residential path
- NAS `proxy-relay` adding its own concurrency queue
- Distributing some sources to off-peak cron slots

### DIFF
--- a/src/app/api/cron/dispatch/route.test.ts
+++ b/src/app/api/cron/dispatch/route.test.ts
@@ -12,7 +12,10 @@ vi.mock("@/lib/db", () => ({
 }));
 vi.mock("@/lib/cron-auth");
 vi.mock("@/lib/qstash");
-vi.mock("@/pipeline/schedule");
+vi.mock("@/pipeline/schedule", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("@/pipeline/schedule")>()),
+  shouldScrape: vi.fn(),
+}));
 
 const APP_URL = "https://hashtracks.com";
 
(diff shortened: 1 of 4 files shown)


johnrclem:
@codex @claude review

PR #955, opened 2026-04-24
perf(hareline): cache event list + tag-invalidate on mutations

johnrclem:
## Summary

- Wrap the slim `/hareline` event list query with `unstable_cache` so anonymous traffic is served from the function's in-region cache instead of hitting Railway on every request.
- Cache key is `(mode, YYYY-MM-DD UTC date)` — rotates at UTC midnight naturally, no live timer.
- `revalidateTag(HARELINE_EVENTS_TAG, "max")` on every mutation that can change what appears on the Hareline page:
  - `scrapeSource()` after merge + reconcile, gated on non-zero delta
  - Admin event mutations (delete, bulk delete, uncancel, backfill city)
  - Admin kennel mutations (update, delete, merge, toggleVisibility) — cache stores denormalized kennel fields per event
  - Admin region mutations (rename, merge, reassign kennels) — `kennel.region` is denormalized
- `revalidate: 3600` fallback for the "scrape silently failed to invalidate" edge case.

## What got dropped

Phase 2b (server-side region scoping for `?regions=` deep links) was explored and reverted after adversarial review surfaced two separate failure modes:

1. **Client/server scope divergence.** `HarelineView.syncUrl()` uses `history.replaceState` for region-chip changes, so Next.js never re-runs the server component. A user who lands on `?regions=state:CA` and then widens their selection would stay frozen on the CA-scoped subset — the client can only filter that subset further, never fetch newly eligible events.
2. **Seed-vs-DB drift.** `expandRegionSelections()` uses `REGION_SEED_DATA` for state/country → metro expansion. Admins create, rename, and reassign regions at runtime via `src/app/admin/regions/actions.ts`, so the seed snapshot misses live metros. `state:CA` deep links would underfetch whenever a California metro was admin-created.

The primary Phase 2 win (cache hits on the DB query + freshness via tag invalidation) is preserved. The dropped optimization was just a share-link payload shrink — modest gain, not worth the correctness risk.

## Correctness

- `dateUtc` serializes to ISO string inside the cache (JSON-safe) and rehydrates to `Date` at the `loadEventsForTimeMode` boundary so consumers (`HarelineView`, `EventCard`, detail panel) see a stable `HarelineListEvent` shape.
- Invalidation coverage expanded beyond the two sites Codex flagged (`toggleKennelVisibility`, `updateKennel`) to include every mutation that touches denormalized fields in the cache: `deleteKennel`, `mergeKennels`, `updateRegionWithRename`, `mergeRegions`, `reassignKennels`.
- `scrape.ts` tag-bust is gated on `created + updated + cancelled + restored > 0` — avoids burning the cache on no-op scrapes.

## Test plan

- [x] `npm run lint` — 0 errors (13 pre-existing warnings in unrelated `travel/*` files)
- [x] `npm test` — 214/214 test files, 5101/5101 tests pass
- [ ] Manual: anonymous `/hareline` — second request noticeably faster than first
- [ ] Manual: trigger a scrape via admin → reload `/hareline` — newly scraped events visible
- [ ] Manual: delete an event via `/admin/events` → `/hareline` drops it immediately
- [ ] Manual: hide a kennel via `/admin/kennels` → its events disappear from `/hareline` without a 1-hour lag
- [ ] Post-deploy: Vercel Speed Insights p50 TTFB drop on `/hareline` for anonymous traffic after cache warm-up

### DIFF
--- a/src/app/admin/events/actions.test.ts
+++ b/src/app/admin/events/actions.test.ts
@@ -24,7 +24,11 @@ vi.mock("@/lib/db", () => ({
     ),
   },
 }));
-vi.mock("next/cache", () => ({ revalidatePath: vi.fn() }));
+vi.mock("next/cache", () => ({
+  revalidatePath: vi.fn(),
+  revalidateTag: vi.fn(),
+  unstable_cache: <T extends (...args: never[]) => unknown>(fn: T) => fn,
+}));
 
 import { getAdminUser } from "@/lib/auth";
 import { prisma } from "@/lib/db";
--- a/src/app/admin/events/actions.ts
+++ b/src/app/admin/events/actions.ts
@@ -3,7 +3,8 @@
 import { prisma } from "@/lib/db";
 import { getAdminUser } from "@/lib/auth";
 import type { ActionResult } from "@/lib/actions";
-import { revalidatePath } from "next/cache";
+import { revalidatePath, revalidateTag } from "next/cache";
+import { HARELINE_EVENTS_TAG } from "@/app/hareline/actions";
 
 const DELETE_BATCH_SIZE = 100;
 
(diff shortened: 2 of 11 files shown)


johnrclem:
@codex @claude review

PR #1076, opened 2026-04-27
fix(adapters/crh3,cah3): drop broken utcRef "Z" suffix

johnrclem:
## Summary

Both Chiang Rai H3 (CRH3) and Cha-Am H3 (CAH3) had an identical broken `utcRef()` helper that mangled timezone-offset publish dates. Empirically discovered via the post-PR #1067 sweep against live sources.

## Root cause

```ts
function utcRef(iso: string | undefined): Date | undefined {
  if (!iso) return undefined;
  return new Date(iso.endsWith("Z") ? iso : iso + "Z");  // ← bug
}
```

Blogger and self-hosted WordPress publish dates carry explicit offsets like `"2026-03-22T18:07:00+07:00"`. The helper appends `"Z"` to that → `"2026-03-22T18:07:00+07:00Z"` → **Invalid Date**.

`chronoParseDate(text, locale, refDate)` then silently fell back to the system clock as reference. With `forwardDate: true`, every year-less post title (e.g. `"CRH3#220 Saturday 26 March"`) failed to resolve.

## Live verification

| Source | Before | After |
|---|---|---|
| Chiang Rai H3 Blog (CRH3) | 3 events, 5 errors | ✅ **6 events, 0 errors** |
| Cha-Am H3 Website (CAH3) | 2 events, 18 errors | unchanged — those posts genuinely have no date in title or body. Separate issue. |

CRH3 is a clean win. CAH3 is unchanged because the failing posts (`"Run 536: Rubber Duck & Paddy Redbelly"`) have no date in either title or body — needs a different fix (publish-date fallback) which is a semantic change worth its own PR + discussion.

## Fix

Drop the bogus suffix and let `new Date(iso)` handle offsets directly. Return undefined on parse failure as a safety net.

## Test plan

- [x] 2 new regression tests (one per adapter) pinning the offset-bearing publish-date format so future utcRef changes can't reintroduce the bug
- [x] `npm test` 5306 pass / 2 skip / 25 todo (was 5304 → +2)
- [x] `npm run lint` 0 errors
- [x] Live-verified CRH3 against production Blogger API: 6 events extracted (was 3), 0 errors (was 5)

## Follow-up

CAH3 still has 18 silent failures — a different bug (no date anywhere in title/body). Possible mitigations: WordPress publish-date fallback, slug-based date extraction, or accept the loss. Will tackle separately.

### DIFF
--- a/prisma/seed-data/sources.ts
+++ b/prisma/seed-data/sources.ts
@@ -81,13 +81,16 @@ function staticScheduleSource(params: {
   url: string;
   kennelTag: string;
   rrule: string;
+  /** Optional anchor for INTERVAL>1 RRULEs (e.g. biweekly) so the
+   *  expansion picks the correct week-of. Format: "YYYY-MM-DD". */
+  anchorDate?: string;
   startTime?: string;
   defaultTitle: string;
   defaultLocation: string;
   defaultDescription: string;
   extra?: Partial<{ trustLevel: number; scrapeFreq: string; scrapeDays: number }>;
 }) {
-  const { name, url, kennelTag, rrule, startTime, defaultTitle, defaultLocation, defaultDescription, extra } = params;
+  const { name, url, kennelTag, rrule, anchorDate, startTime, defaultTitle, defaultLocation, defaultDescription, extra } = params;
   return {
     name,
     url,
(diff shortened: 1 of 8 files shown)


johnrclem:
@claude @codex @gemin review

PR #1098, opened 2026-04-28
fix(scrape): centralize safeRevalidateTag for CLI-safe cache busts

johnrclem:
## Summary

Extracts the inline try/catch around `revalidateTag(HARELINE_EVENTS_TAG)` in `scrapeSource()` into a reusable `safeRevalidateTag` helper in [src/lib/safe-revalidate.ts](src/lib/safe-revalidate.ts). The helper swallows the specific Next.js E263 "outside request scope" invariant (so CLI/tsx scripts no longer flip a successful scrape to `success: false`) and re-throws every other error so production cache-busting bugs still surface.

- Filter is narrow: matches `__NEXT_ERROR_CODE === "E263"` first, then a start-anchored regex that *requires* the `in revalidateTag ` / `in revalidatePath ` suffix — a future same-prefix-different-suffix invariant won't be silently masked.
- Logs the suppressed `Error` object in full on the swallow path, not just the message.
- Exports `safeRevalidatePath` with the same contract for future callers; no current call sites change.

**Production cache-busting is unaffected.** QStash/cron scrapes still run inside a request scope, so `revalidateTag` fires normally and the Hareline `unstable_cache` is busted on the same trigger as before.

## Repro

Before:
```sh
set -a; source .env; set +a
# any tsx script calling scrapeSource(...) directly
npx tsx scripts/repro-scrape.ts
# → { success: false, errors: ["Invariant: static generation store missing in revalidateTag hareline:events"], ... }
```

After:
```
[safe-revalidate] revalidateTag(hareline:events) skipped: no request scope <Error>
{ "success": true, "eventsFound": 25, "updated": 25, "errors": [] }
```

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm run lint` (0 errors; pre-existing warnings only)
- [x] `npm test` — 5366 passed, 0 failed (10 new tests in `safe-revalidate.test.ts`, including a regression for "same-prefix lookalike must re-throw")
- [x] CLI repro against live `DeMon H3 Google Calendar` — `success: true`, 25 events written
- [x] `/codex:adversarial-review` — addressed the medium finding by tightening the regex from prefix-only to suffix-anchored, and by including the original Error in the warn log
- [x] `/simplify` — added file-level JSDoc (matching `src/lib/indexnow.ts` convention), trimmed redundant comment in `scrape.ts`, gave the test spies proper `beforeAll`/`afterAll` lifecycle

### DIFF
--- a/src/lib/safe-revalidate.test.ts
+++ b/src/lib/safe-revalidate.test.ts
@@ -1,0 +1,169 @@
+import { describe, it, expect, vi, beforeEach, beforeAll, afterEach, afterAll } from "vitest";
+import type { MockInstance } from "vitest";
+
+vi.mock("next/cache", () => ({
+  revalidateTag: vi.fn(),
+  revalidatePath: vi.fn(),
+}));
+
+import { revalidateTag, revalidatePath } from "next/cache";
+import { safeRevalidateTag, safeRevalidatePath } from "./safe-revalidate";
+
+const mockRevalidateTag = vi.mocked(revalidateTag);
+const mockRevalidatePath = vi.mocked(revalidatePath);
+
+function makeStaticStoreError(
+  expression: string,
+  opts: { withCode?: boolean } = { withCode: true },
+): Error {
+  const err = new Error(`Invariant: static generation store missing in ${expression}`);
+  if (opts.withCode) {
+    Object.defineProperty(err, "__NEXT_ERROR_CODE", {
+      value: "E263",
+      enumerable: false,
+      configurable: true,
+    });
+  }
+  return err;
+}
+
+describe("safeRevalidateTag", () => {
+  let consoleWarnSpy: MockInstance<typeof console.warn>;
+
+  beforeAll(() => {
+    consoleWarnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+  });
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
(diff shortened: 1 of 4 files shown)


johnrclem:
@codex @claude review

johnrclem:
Addressed all three review findings in [`ebaace84`](https://github.com/johnrclem/hashtracks-web/pull/1098/commits/ebaace84):

1. **Best-effort regression** — restored the outer try/catch around `safeRevalidateTag(...)` at [src/pipeline/scrape.ts:514](src/pipeline/scrape.ts:514) using `logHousekeepingError`, so any unexpected non-E263 revalidation failure still doesn't roll back a successful scrape. The helper keeps its narrow E263-only swallow contract.
2. **`profile` optional** — defaulted to `{ expire: 0 }` (canonical "invalidate now"). Also avoids the Next.js deprecation warning on the no-arg path.
3. **Test duplication (SonarCloud gate)** — lifted the `consoleWarnSpy` lifecycle (`beforeAll`/`beforeEach`/`afterEach`/`afterAll`) to module scope so both `describe` blocks share it. Clears the 15.7% duplication finding.

Bonus: typed `safeRevalidatePath`'s `type` as `Parameters<typeof revalidatePath>[1]` for parity with `profile`.

Re-verified: `tsc --noEmit` ✓, `lint` (0 errors) ✓, `npm test` (29 of 5366 in scope) ✓, live CLI repro against `DeMon H3 Google Calendar` → `success: true`, 25 events written ✓.

PR #1092, opened 2026-04-28
feat(schema): add EventKennel join table + backfill (step 1 of #1023)

johnrclem:
## Summary

First implementation step of the multi-kennel co-hosted events spec ([docs/multi-kennel-events-spec.md](docs/multi-kennel-events-spec.md), merged via [#1071](https://github.com/johnrclem/hashtracks-web/pull/1071)). **Purely additive** — no application code changes. Step 2 (dual-write across all Event-write sites) follows in a separate PR.

### What lands

- **`EventKennel` model** in `prisma/schema.prisma` — `(eventId, kennelId, isPrimary)`, composite PK, index on `kennelId`, FK cascade on Event delete. Reverse relations on `Event.kennels` and `Kennel.eventKennels`.
- **Two migrations** (hand-written; pre-existing checksum drift on `20260409000000_add_audit_issue_mirror` blocks `prisma migrate dev`):
  - `add_event_kennel_join_table` — CREATE TABLE + indexes + FKs. SQL matches what Prisma 7 would generate (verified by inspecting the applied table).
  - `backfill_event_kennel_from_kennel_id` — INSERT…SELECT one EventKennel row per existing Event with `isPrimary=true`, `ON CONFLICT DO UPDATE SET isPrimary=true` (spec D17). Then `CREATE UNIQUE INDEX (eventId) WHERE isPrimary=true` enforcing the single-primary invariant at the DB level (spec D13).
- **`scripts/verify-event-kennel-backfill.ts`** — post-deploy verification asserting four invariants: total ≥ event count, exactly one primary per event, partial unique index exists, primary `EventKennel.kennelId` matches `Event.kennelId`. Exits non-zero on failure so on-call / CI can gate on it.

### Why hand-written migrations

`prisma migrate dev` errored on a pre-existing checksum drift in the migration history (`20260409000000_add_audit_issue_mirror` was modified post-apply). Resetting the local DB to recover would lose data; the migrations here are purely additive and the SQL matches Prisma's standard generation pattern (TEXT for String, BOOLEAN NOT NULL DEFAULT false, `<Table>_pkey`/`<Table>_<col>_idx`/`<Table>_<col>_fkey` naming, ON DELETE CASCADE/RESTRICT). `migrate deploy` applies both successfully.

### Local verification

Refreshed `hashtracks_dev` from prod (34,205 events), applied both migrations, ran the verification script:

```
Event count:           34,205
EventKennel total:     34,205
EventKennel primaries: 34,205
Partial unique index:  present
[...]

Issue #1023, opened 2026-04-26
Schema: support kennels co-hosting a single event (multi-kennel events)


johnrclem:
## Context

HashTracks currently models `Event.kennelId` as a single foreign key ([prisma/schema.prisma:307](prisma/schema.prisma)). When two kennels jointly host one real-world trail, our adapters route the event to a single kennel's page and the other kennel's members never see it on their kennel page.

Surfaced concretely by:
- **#991** (Cherry City / OH3 joint inaugural) — the surgical fix is to route to cch3-or via kennelPattern order, but OH3 members miss the event on the OH3 page.
- Likely recurring pattern for anniversary trails, regional inter-kennel events, "joint" weekend bashes.

Closed routing-only issues (#938, #850, #757, #670, #817) were all solved by single-kennel routing — they were attribution bugs, not true co-hosts. #991 is the first genuine co-host case.

## Design questions to resolve

- **Schema:** new `EventKennel` join table vs. `Event.kennelIds: String[]`? Primary vs. secondary distinction? How does this interact with the `(kennelId, date)` dedup key?
- **Adapter signal:** how does an adapter emit "this event has two kennels"? Extend `RawEventData.kennelTag` to `kennelTags: string[]`? New `coHostKennelTags`?
- **Dedup semantics:** if cch3-or's own website *also* publishes this trail, the merge pipeline needs to recognize "this Cherry City+OH3 event from Oregon Calendar" and "this Cherry City event from cherry-city.org" as the same row. Cross-source merge across overlapping kennel sets, not strictly identical sets.
- **Display layer:**
  - Kennel pages — does this event appear on both kennel pages? With what visual treatment?
  - Event detail page — show both kennel badges, both calendars/ICS feeds.
  - Hareline calendar — one card or two?
  - Misman attendance — which kennel's roster owns the attendance record?
  - Hash House Lookup / kennel directory — count this event toward both kennels' stats?
- **Filtering & queries:** every `where: { kennelId }` filter (server actions, kennel page, subscriptions, near-me) needs to switch to "kennel is in event's kennel set".
- **Migration:** ~50k existing events to backfill (single-kennel → array of one). Reversible.
- **kennelPattern adapter config:** today `[regex, kennelTag]` — would need to allow `[regex, kennelTag[]]` for true co-host detection like `Cherry City.*OH3 → [cch3-or, oh3]`.

## Acceptance criteria (when this gets picked up)

[...]

PR #1099, opened 2026-04-28
feat(dual-write): write EventKennel alongside Event creates (step 2 of #1023)

johnrclem:
## Summary

Second implementation step of the multi-kennel co-hosted events spec ([docs/multi-kennel-events-spec.md](docs/multi-kennel-events-spec.md)). Step 1 (schema + backfill) landed in [#1092](https://github.com/johnrclem/hashtracks-web/pull/1092). This PR makes every Event-write site also write the matching primary EventKennel row in the same transaction, and updates admin kennel-merge / cascade-delete to handle the join table.

### What lands

**`src/lib/event-write.ts` (new)** — `createEventWithKennel(client, data)` helper. Uses Prisma's nested write so Event + primary EventKennel land in one round-trip with implicit transaction. Input type narrowed to scalars only (`Omit` Prisma's nested relation creates) so no caller can smuggle extra primary rows that bypass the helper's invariant.

**Call sites:**
- `pipeline/merge.ts:1170` — replaces the bare `prisma.event.create` with the helper. Hot path: scrape cron creates 100s of events per run; nested write keeps per-event cost flat.
- `app/logbook/actions.ts` manual entry — uses the helper inside the existing `$transaction` so Event + EventKennel + Attendance are all atomic.

**`src/app/admin/kennels/actions.ts`:**
- `deleteKennel` — new co-host guard counts `EventKennel` rows for this kennel; refuses delete with a friendly error if non-zero. Without it, the FK RESTRICT on `EventKennel.kennelId` would surface as a confusing FK violation deep in the cascade.
- `mergeKennels` — new `deduplicateEventKennels` helper handles the four collapse/re-point cases. Wraps dedup + reassignment + delete in one interactive `prisma.$transaction(async (tx) => {...})`. Concurrent `Event/EventKennel` inserts that race the merge either commit before our snapshot (we re-point them correctly) or after our delete (FK RESTRICT cleanly aborts only the racing tx).

### Codex adversarial review

Codex caught **3 BLOCKERs** before this PR went up — all addressed:
1. **Collapse-order partial-index conflict** — original code promoted target to `isPrimary=true` while source was still primary, violating the partial unique index. Fixed: delete source FIRST, then promote target.
2. **Dedup running outside the merge tx** — could leave the denorm/join sync invariant violated mid-merge. Fixed: dedup now runs inside the same interactive transaction as the reassignment + delete.
3. **Race with concurrent creates** — a scrape running mid-merge could create `EventKennel(source)` after dedup finished, leaving stale rows. Fixed: in-tx + FK-RESTRICT-on-commit handle this cleanly (either we abort, or the racing tx does).

Plus several IMPORTANT findings: helper input type narrowed, misleading comment fixed, four collapse cases now have unit tests, live verification script added that exercises all four cases against the actual partial unique index.

### Local verification (hashtracks_dev, 34,205 events)

- 5,368 unit tests pass (304 across the 3 affected test files); `tsc --noEmit` clean.
- Lint shows only pre-existing warnings unrelated to this diff.
- `scripts/live-verify-dual-write.ts`: dual-write succeeds; composite PK rejects duplicate; partial unique index rejects second `isPrimary=true` row.
- `scripts/live-verify-event-kennel-dedup.ts`: all 4 collapse cases + multi-row case verified live (the partial unique index actually fires at the right transition points).

### Production safety

- `pipeline/merge.ts` hot path: per-event cost is one round-trip (nested write) inside Prisma's implicit transaction.
- Admin merge: single interactive transaction. Concurrent scrape creates either commit before our snapshot (handled by dedup) or after our commit (FK RESTRICT cleanly fails their tx, our merge stays consistent).
- `deleteKennel` guard prevents the FK-violation-as-cascade-error UX problem.

### What's NOT in this PR

- Display-layer migration (~200 REWRITE sites) — that's step 5 in the spec.
- Adapter `kennelTag` → `kennelTags: string[]` codemod — that's step 3.
- `Event.kennelId` removal — step 7, after a 2-week soak post-step-5.

## Test plan

- [x] `npm run lint` — only pre-existing warnings
- [x] `npm test` — 5368 pass, no new failures
- [x] `scripts/verify-event-kennel-backfill.ts` still passes (no regression on step 1's invariants)
- [x] `scripts/live-verify-dual-write.ts` — 3 invariants pass
- [x] `scripts/live-verify-event-kennel-dedup.ts` — 4 collapse cases + multi-row case pass
- [ ] Vercel preview deploys cleanly
- [ ] Post-merge: monitor scrape error rates for ~24h to confirm no per-event regression in the merge pipeline

Refs #1023

### DIFF
--- a/scripts/live-verify-dual-write.ts
+++ b/scripts/live-verify-dual-write.ts
@@ -1,0 +1,108 @@
+/**
+ * One-shot manual verification (#1023 step 2): exercises createEventWithKennel
+ * against hashtracks_dev to confirm the partial unique index enforces the
+ * single-primary invariant and that the helper writes both rows atomically.
+ *
+ * Run: `npx tsx scripts/live-verify-dual-write.ts`
+ *
+ * NOT a vitest fixture — this is an interactive smoke test for the dev DB.
+ * Idempotent (cleans up rows it creates).
+ */
+import "dotenv/config";
+import { Prisma } from "@/generated/prisma/client";
+import { prisma } from "@/lib/db";
+import { createEventWithKennel } from "@/lib/event-write";
+
+async function main() {
+  const kennel = await prisma.kennel.findFirst({ select: { id: true, shortName: true } });
+  if (!kennel) throw new Error("No kennels in hashtracks_dev — bail");
+
+  const baseDate = new Date(Date.UTC(2099, 0, 1, 12));
+
+  console.log(`Using kennel: ${kennel.shortName} (${kennel.id})`);
+
+  const event = await prisma.$transaction((tx) =>
+    createEventWithKennel(tx, {
+      kennelId: kennel.id,
+      date: baseDate,
+      trustLevel: 5,
+      title: "DUAL_WRITE_PROBE",
+    }),
+  );
+  console.log(`Created Event ${event.id} via dual-write`);
+
+  const ek = await prisma.eventKennel.findUnique({
+    where: { eventId_kennelId: { eventId: event.id, kennelId: kennel.id } },
+  });
+  if (!ek || !ek.isPrimary) {
+    throw new Error(`Primary EventKennel row missing or not primary for ${event.id}`);
+  }
+  console.log("Primary EventKennel row written and isPrimary=true ✓");
+
+  // Both checks below must surface as Prisma P2002 (Unique constraint failed).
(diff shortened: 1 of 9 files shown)


johnrclem:
@claude @codex review

PR #1181, opened 2026-05-01
feat(audit): chrome-stream filing endpoints with payload-bound nonces (5b)

johnrclem:
## Summary

Adds the chrome-stream filing flow that earlier bundles set up the foundation for. Chrome agents running in an admin's browser session can now file audit findings through a structured server endpoint instead of opening GitHub issue-new URLs by hand. The server validates the payload, computes the fingerprint from the rule registry, and either coalesces against an existing AuditIssue with the same fingerprint OR files a fresh GitHub issue with the canonical block embedded for sync mirroring.

Builds on PRs #1162, #1163, #1165, #1171, #1172.

## Endpoints

**`POST /api/audit/mint-filing-nonce`** — admin-auth'd, returns a single-use nonce bound to `(adminUserId, kennelCode, ruleSlug, payloadHash)`. Origin check fails closed before DB. Raw nonce returned to agent; only sha256 hash persisted (`AuditFilingNonce.nonceHash`).

**`POST /api/audit/file-finding`** — validates Origin + admin session + all nonce binds. Coalesces against an existing fingerprint match (comments on it) or files a fresh GitHub issue with the canonical block embedded.

## Codex pass-2 review surfaced 3 real correctness issues, all fixed pre-push

| Severity | Finding | Resolution |
|---|---|---|
| HIGH | `title` not hash-bound — caller could change title post-mint | Added `title` to `FilingPayload`; bound in `computePayloadHash` |
| HIGH | Failed coalesce-comment fell through to fork a duplicate issue | Returns 502 with `existingIssueNumber`; no fork |
| HIGH | Nonce consumed before GitHub call → no retry on transient failure | New `AuditFilingNonce.filingResultJson` column caches outcome; retries with same nonce return cached result |

## Schema

`AuditFilingNonce` gains `filingResultJson Json?` for the idempotency cache. Migration `20260501112723_audit_filing_nonce_result` applied locally; Vercel runs `prisma migrate deploy` on build.

**Trade-off documented**: two concurrent requests with the same nonce can each pass bind checks and reach the GitHub call, producing one orphan issue. Acceptable for admin-driven audit-filing flow (low concurrency); outbox/job model is the long-term answer if parallelism becomes a problem.

## `/simplify` cleanup applied pre-PR

Three actionable findings from the parallel review agents:
- Collapsed `updateMany` + `findUnique` consume into a single `findUnique` + `update` pattern (payload-tamper detection now lives in explicit code, not stringly-typed WHERE clauses)
[...]

PR #1198, opened 2026-05-01
feat(audit): chrome prompts call file-finding endpoint (5c-C)

johnrclem:
## Summary

Wires the deep-dive (chrome-kennel) and hareline (chrome-event) prompts to the audit API endpoints shipped in #1181/5c-A so chrome agents file findings via the same fingerprint-aware path that the cron uses, instead of the legacy GitHub-issue-URL prefill. **This is what activates 5a/5b/5c-A end-to-end.**

## Endpoint extension

`POST /api/audit/mint-filing-nonce` now accepts a canonical-fields shape in addition to the existing pre-hashed shape:

```json
{
  "stream": "CHROME_KENNEL",
  "kennelCode": "nych3",
  "ruleSlug": "hare-url",
  "title": "...",
  "eventIds": ["..."],
  "bodyMarkdown": "..."
}
```

Server computes the payload hash via `computePayloadHash` (same helper `file-finding` uses to verify on consume), so chrome agents don't need to touch `crypto.subtle`. Persisted hash is identical regardless of shape, so the consume-side recompute-and-verify works end-to-end.

## Prompt rewrite

`renderFilingInstructions` now surfaces three options:

| Option | Use | Why |
|---|---|---|
| **1 — Audit API (preferred)** | Two-step `mint-filing-nonce` → `file-finding` flow with example JSON payloads + response shapes documented inline | Cross-stream coalescing fires immediately; recur outcomes prevent duplicates |
| **2 — Paste finding** | Fallback when API call fails after a fresh nonce | Unchanged from before |
| **3 — Legacy GitHub URL** | Last-resort bridging | Bridging tier in 5c-A picks these up on next sync |

## Codex pass-1 catch

Earlier revision emitted invalid JSON in the hareline prompt because `{KENNEL_CODE}` was left unquoted. Fixed: always JSON-quote the kennelCode value, even the literal placeholder. **New regression test** extracts every \`\`\`json fenced block from the prompt and runs `JSON.parse` on each.

## Test plan

- [ ] Local: ✅ 5776 tests pass, tsc clean, lint clean
- [ ] SonarCloud / Codacy clean
- [ ] Manual: copy chrome-event prompt → paste into Claude in Chrome → agent finds an event-quality issue → calls `mint-filing-nonce` with canonical fields → calls `file-finding` → issue appears in GH with correct labels
- [ ] Manual: file the same finding twice → second call returns `recurred` (strict tier), `recurrenceCount: 2`, comments "Still recurring on …" on the existing issue
- [ ] Manual: hareline prompt's JSON examples are syntactically valid

### DIFF
--- a/src/app/api/audit/mint-filing-nonce/route.test.ts
+++ b/src/app/api/audit/mint-filing-nonce/route.test.ts
@@ -10,6 +10,7 @@ vi.mock("@/lib/site-url", () => ({
 
 import { getAdminUser } from "@/lib/auth";
 import { prisma } from "@/lib/db";
+import { computePayloadHash, type FilingPayload } from "@/lib/audit-nonce";
 import { buildApiPostRequest, type ApiPostInit } from "@/test/audit-request";
 import { POST } from "./route";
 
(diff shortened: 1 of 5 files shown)


johnrclem:
@claude review

PR #1203, opened 2026-05-01
feat(audit): server-side queue-snapshot token for deep-dive completion (closes #1160)

### LINKED ISSUE #1160: Admin audit — Mark deep dive complete dropdown is unsafe for Claude/Chrome automation (GGFM marked as Galveston)
## Summary

While completing the GGFM deep dive (issue #1159), I (Claude in Chrome) submitted the second mark-complete attempt against **Galveston** instead of GGFM. The user caught this visually after the fact. This is the latest in a recurring pattern of admin-audit dialog UX issues that make this surface fragile for automated agents.

## What happened (this incident)

1. Filed finding #1159 against GGFM.
2. Navigated to /admin/audit, opened the "Deep dive target" combobox, selected GGFM, clicked Mark complete, JS-submitted the dialog form. Coverage advanced 101 → 102, but a hard reload showed GGFM was still in the queue (the well-known intermittent "first submit doesn't drop the kennel" flake).
3. Re-opened the dropdown and ran a `find` for "GGFM option". The accessibility tree returned a ref labeled "GGFM option (combobox listbox option, location New York City, NY)". I clicked that ref.
4. Clicked Mark complete, JS-submitted. Coverage advanced 102 → 103. **GGFM was still in the queue afterward**, which I attributed to the same flake — but the user observed that the dropdown displayed Galveston when I submitted, not GGFM. So that submit credit went to Galveston, not GGFM.

The most likely root cause: between the first dialog close and the second dropdown open, the queue list re-rendered (because GGFM had moved or rendering shifted around it). The DOM node that had been "GGFM option" got recycled/repositioned, but my snapshot's ref still pointed to a DOM position that now held Galveston. The ref's accessibility-tree label was stale relative to the DOM at click time.

## Why this is a Claude-in-Chrome UX problem (not just operator error)

The current mark-complete UX has several properties that make it brittle for automated agents:

1. **Combobox uses position-based options with no stable identifier.** When the queue re-renders, the same option can refer to a different kennel. Refs from `find` snapshots become stale silently — there's no exception, just a wrong submit.

2. **The selected target is not echoed inside the confirmation dialog.** The dialog shows the count and summary fields, but not the kennel name being marked complete. So there's no in-dialog last-chance verification — the agent (and the human) has to trust the dropdown state shown on the page behind the dialog.

3. **The dialog form has no kennel-id hidden input that's visible to the agent.** I can't post-hoc verify which kennel my submit will be applied to before clicking Submit, except by reading the dropdown label, which is exactly the value that just got stale.

4. **Coverage counter increments even when the named kennel doesn't drop from the queue.** This is the documented intermittent flake (~25% need a single submit, ~75% need two). It trains agents to keep re-submitting until the queue updates — which is precisely the loop in which this misattribution happened.

5. **No undo, no audit log surfaced in the UI.** The "Recent audit runs (last 14 days)" panel shows aggregate event counts, not per-kennel deep-dive completions. So the agent can't verify after the fact which kennel actually got credited; only the human can (by spotting it in the dropdown at submit time, as the user did here).

## Concrete asks (in priority order)

1. **Show the target kennel inside the mark-complete dialog header.** "Mark deep dive complete: **GGFM** (New York City, NY) — 1 finding". This single change would have caught this incident.

2. **Make the mark-complete form take a kennel slug, not a position-based dropdown index.** Backend should reject the submit if the slug doesn't match the kennel currently shown in the queue at that position, OR (better) make the dialog hidden-input the slug at dialog-open time and not re-derive from the dropdown.

3. **Surface a "last 20 deep dives completed" panel** with timestamp + kennel + finding count + summary, so misattributions are visible to humans during review and to agents during verification.

4. **Fix the intermittent first-submit flake.** Whatever bookkeeping is happening between "increment coverage" and "remove from queue" is racy and is the trigger for the double-submit loop that caused this misattribution. If those two state changes were atomic (or at least both verified before dialog close), agents wouldn't be trained to re-submit.

5. **Optional but nice: add an idempotency token / dedupe.** If the same kennel slug is marked complete twice within 60s, second one should be a no-op rather than incrementing a different kennel's counter.

## Cleanup needed

- **Galveston** (`gal`, Los Angeles CA) — has incorrectly received a deep-dive completion credit. Coverage 102 → 103 of that submit should be reverted from Galveston, and Galveston should be returned to the queue. (Galveston was NOT actually audited.)
- **GGFM** still needs its mark-complete done correctly. As of writing, it's still in the queue at line 163 of the queue table.

I am pausing on touching the queue further until you confirm the cleanup approach, so I don't compound the problem.

## Cross-references

- The GGFM deep-dive findings issue: #1159
- Prior offer to file an admin-tooling issue about mark-complete persistence flakiness was made in two previous sessions and not actioned at the time. This incident makes that escalation overdue.

johnrclem:
## Summary

Closes #1160. The "Mark deep dive complete" dialog was misattributing ~75% of clicks because the dropdown re-derived the target kennel at submit time. A queue change between dialog-open and submit silently shifted which kennel got credit — the original report had Galveston marked complete after a click on GGFM.

**Fix**: bind the kennel at click-time via an HMAC-signed token capturing `(kennelCode, queueSnapshotId, expiresAt)`. Server recomputes the snapshot from the live queue at submit time and rejects mismatches with a typed error so the dialog can branch.

## Surface

| Layer | Change |
|---|---|
| **`src/lib/queue-snapshot-token.ts`** | Pure HMAC helpers — `computeQueueSnapshotId` (SHA-256 over sorted kennelCodes, stable across reorders), `signQueueToken`/`verifyQueueToken` (HMAC-SHA256 + 10-min TTL + tagged-result verifier) |
| **`getDeepDiveQueueToken(kennelCode)`** | New admin action; mints a token if the kennel is in the current queue, returns null otherwise |
| **`recordDeepDive`** | Now requires `queueToken`. Returns `RecordDeepDiveResult` discriminated union: `{ ok, id }` / `invalidToken` / `queueChanged` / `kennelGone` |
| **`DeepDiveCompleteDialog`** | Fetches token on open; on `queueChanged` auto-refetches + shows "queue was edited, confirm by submitting again"; on `kennelGone`/`invalidToken` shows hard-error message; submit disabled when token isn't loaded; title echoes `${shortName} (${region})` |
| **`AUDIT_QUEUE_TOKEN_SECRET`** | New env var documented in CLAUDE.md |

## Codex pass-1 catch

`Buffer.from(macHex, "hex")` silently truncates on non-hex input, which would crash `timingSafeEqual` on a same-length non-hex MAC — turning attacker-controlled input into a 500 path. Pre-validate hex shape upfront, returning `invalid-signature`. Regression test included.

## Test plan

- [ ] Local CI gate: ✅ 5794 tests pass, tsc clean, lint clean. 15 token-helper tests, 9 recordDeepDive tests, 3 getDeepDiveQueueToken tests.
- [ ] SonarCloud / Codacy clean
- [ ] **Set `AUDIT_QUEUE_TOKEN_SECRET` in Vercel before merge** — `openssl rand -hex 32` is fine. The action throws if the secret is missing.
- [ ] Manual: open the deep-dive dialog → check that the title shows kennel + region → submit happy path → log row created
- [ ] Manual: from a second tab, edit the queue (e.g. mark another kennel complete) → submit the first dialog → expect `queueChanged` warning → submit again → success
- [ ] Manual: from a second tab, mark THIS kennel complete via the API → submit the first dialog → expect `kennelGone` hard-error
- [ ] Manual: tamper with the token in DevTools → submit → expect `invalidToken` hard-error

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,6 +66,7 @@ logbook + kennel directory.
 - NEXT_PUBLIC_CLERK_SIGN_UP_URL=/sign-up
 - GEMINI_API_KEY=         # Google AI API key (Sprint 10+)
 - CRON_SECRET=            # Secret for cron auth (Bearer token fallback)
+- AUDIT_QUEUE_TOKEN_SECRET= # HMAC secret for deep-dive queue-snapshot tokens (issue #1160). 32+ random bytes; rotate independently from CRON_SECRET.
 - QSTASH_TOKEN=           # Upstash QStash API token (fan-out job dispatch)
 - QSTASH_CURRENT_SIGNING_KEY= # QStash signature verification (current key)
 - QSTASH_NEXT_SIGNING_KEY=    # QStash signature verification (next key for rotation)
--- a/src/app/admin/audit/actions.test.ts
+++ b/src/app/admin/audit/actions.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from "vitest";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 
 vi.mock("@/lib/auth", () => ({ getAdminUser: vi.fn() }));
 vi.mock("@/lib/db", () => ({
@@ -63,8 +63,14 @@ import {
   getDeepDiveQueue,
   getDeepDiveCoverage,
   getCloseReasonRatiosByStream,
+  getDeepDiveQueueToken,
   recordDeepDive,
 } from "./actions";
+import {
+  computeQueueSnapshotId,
+  computeQueueTokenExpiresAt,
+  signQueueToken,
+} from "@/lib/queue-snapshot-token";
 
 const mockAdmin = vi.mocked(getAdminUser);
 const mockLogFind = vi.mocked(prisma.auditLog.findMany);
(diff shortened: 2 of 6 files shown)


johnrclem:
@claude review

PR #1264, opened 2026-05-06
fix(audit): pre-mint deep-dive tokens at page render (closes #1207, #1216, #1175)

### LINKED ISSUE #1175: Admin audit — Deep dive target combobox returns wrong option when selected by a11y ref (repro of #1160, caught pre-submit)
## Summary

Direct reproduction of #1160. While trying to mark Glasgow H3 deep dive complete, the `find` tool returned a ref labeled `option "Glasgow H3"` (ref_464) — but clicking it actually selected **Giggity** in the combobox. **Caught pre-submit this time** thanks to the new guard rule in the kennel-prompt template, so no misattribution actually shipped (Giggity has NOT been incorrectly credited).

## Reproduction (this turn)

1. Navigated to https://hashtracks.xyz/admin/audit (post-#1160).
2. `find("deep dive target combobox dropdown")` → returned `ref_229`. Clicked it. Listbox opened.
3. `find("Glasgow H3 option in deep dive target listbox")` → returned `ref_464` with description `option "Glasgow H3"` and "This is the Glasgow H3 option element within the listbox [ref_420]". Clicked it.
4. **Verification step (new) — JS-read the combobox display text before clicking Mark complete:**

   ```js
   document.querySelector('[role="combobox"]').innerText
   // Returns: "Giggity\nBremerton, WA"
   ```

5. Combobox displayed Giggity, not Glasgow H3. Stopped immediately, did not click Mark complete.

## Root cause hypothesis

The accessibility tree (what `find` reads from) is being computed from a DOM snapshot that includes options 25–30 positions deep into the listbox, whereas only ~12 options are actually present in the visible/scrolled-into-view virtualized region. When `find` returns a ref for an option index past the rendered window, the click event lands on a **different** rendered option that happens to occupy that DOM position.

This is the same class of bug as #1160 but with a different surface — last time the queue had re-rendered between snapshot and click; this time the listbox itself is virtualized/windowed so most "options" in the a11y tree don't have stable DOM nodes at the time of click.

## Why the previous "always re-find immediately before click" mitigation was insufficient

I followed the agreed mitigation from #1160:
- Re-found the dropdown ref immediately before opening it ✓
- Re-found the option ref immediately before clicking it ✓
- Did NOT reuse refs across re-renders ✓

**It still failed.** The ref returned was wrong on its very first read.

The verification step that actually caught it was reading `[role="combobox"].innerText` AFTER click, BEFORE submit. **That is now mandatory** for any agent using this dialog.

## Concrete asks (additive to #1160)

In addition to the asks already in #1160:

6. **Make every option in the deep-dive-target combobox have a stable `data-kennel-slug` attribute** that the agent can read AFTER click to verify the selection landed on the intended kennel. The `innerText` read works but is fragile — a slug attribute on the combobox after selection (`data-selected-kennel="glasgow-h3"`) would be much more reliable than parsing the display text.

7. **Either fix the listbox virtualization** so all rendered options in the a11y tree correspond to clickable DOM nodes, **or** disable virtualization entirely on this combobox — the kennel list is bounded (~342 entries) and not so large that virtualization is necessary.

8. **Add a server-side guard** that rejects mark-complete submissions where the submitted kennel slug does not match the slug currently shown in the queue's "next up" position. This catches misattribution attempts even when the UI fails.

## Side note on the `find` tool's own contract

The `find` tool description for this Chrome extension says it returns "matching elements" — but in this case the element it returned, while semantically labeled "Glasgow H3 option", was not the element that would receive the click. From the agent's side this looks like a `find`-tool-vs-DOM-truth mismatch on virtualized lists. Worth mentioning to the Chrome-extension team if appropriate; the kennel-page-side fix is still the right one regardless.

## Status of Glasgow H3 deep dive

* Findings filed: #1173 (profile bundle), #1174 (#2213 missing Location).
* **Not yet marked complete on the dashboard** — pausing per the kennel prompt's guard rule.
* Coverage will need to be reconciled by an admin once the combobox issue is mitigated, OR the agent can retry using an alternative selection mechanism (e.g. typing the kennel name into a search field if one exists, or scrolling the listbox so that the target is in the rendered window before reading the ref). Awaiting direction.

## Cross-references

- #1160 — original misattribution (GGFM marked as Galveston) — same root cause class, this is a fresh repro
- #1173, #1174 — the Glasgow H3 findings that are blocked behind this

### LINKED ISSUE #1207: Admin audit: Mark deep dive complete dialog showed Server Components error and submit did not persist for GLH3
## Symptom

After completing the GLH3 deep dive, opened the Mark deep dive complete dialog (header confirmed "Mark deep dive complete: GLH3 (Lansing, MI)" and submit button "Mark GLH3 complete"). The dialog rendered with a visible error block underneath the form fields:

> "An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this error instance which may provide additional details about the nature of the error."

I filled in:
- Findings filed: 1
- One-line summary: Logo URL missing — kennel website (glh3.net) exposes a stable Squarespace CDN logo image; rest of profile and event extraction (16 events from Google Calendar) verified accurate.

…and clicked **Mark GLH3 complete**.

## Result

After hard-reload of `/admin/audit`:
- GLH3 still appears in the queue with `last dived: never`
- Coverage remained at `114 / 341 active kennels (33%)` — i.e. unchanged, no kennel was marked complete by this submit (not even a misattribution to a different kennel as in #1160)

So this looks like a **submit-failure** mode rather than the misattribution mode in #1160. The dialog rendered with a Server Components error already visible *before* I clicked Mark complete, which suggests something on the server-side is throwing during the Mark-complete dialog render or the action handler.

## Per the post-submit guard

Per the audit prompt's guard rule (referencing #1160), I have **not retried the submit**. Filing this issue and leaving GLH3 in the queue for an admin to investigate.

## Findings

The actual finding for GLH3 (#1206 — kennel-profile-logo-missing) was filed successfully via the new `/api/audit/file-finding` flow before the failed mark-complete submit. So no findings work was lost; just the deep-dive bookkeeping didn't persist.

## Suggested next step

- Inspect server logs around the Mark-complete dialog for GLH3 (the "digest" property of the React error should map to the underlying exception).
- Once root cause is fixed, an admin can mark GLH3 complete on my behalf so it drops from the queue (1 finding filed, summary above).

## Stream / rule

`stream: CHROME_KENNEL`, `rule: admin-tooling-deep-dive-submit-failure`

### LINKED ISSUE #1216: Admin audit: Mark deep dive complete failed for GGFM (Server Components render error, submit not persisted)
## Recurrence of #1207

Same failure mode as #1207 (filed for GLH3 earlier this session) reproduced for **GGFM**.

## Symptom

- Selected GGFM via queue-row click. Combobox header read "GGFM, New York City, NY · 1 source · 12 events in last 90d · last dived never".
- Opened the Mark deep dive complete dialog. Header confirmed: "Mark deep dive complete: GGFM (New York City, NY)". Submit button: "Mark GGFM complete".
- Dialog rendered with a visible error block underneath the form fields:
  > "An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this error instance which may provide additional details about the nature of the error."
- I filled in: Findings filed = 0, One-line summary referencing the existing #1159 finding.
- Clicked "Mark GGFM complete".

## Result

After hard-reload of `/admin/audit`:
- GGFM still appears in the queue with `last dived: never`
- Coverage stayed at `114 / 341 active kennels (33%)` — unchanged.

Same failure shape as #1207 (GLH3): the dialog renders with the Server Components error already visible, and submit is a no-op (no kennel marked, not even misattributed elsewhere).

## Per the post-submit guard

I did not retry the submit. Filing this issue and leaving GGFM in the queue.

## Findings recap for GGFM

GGFM has **0 new findings** from this deep-dive round — the only outstanding profile gap (dedicated GGFM Facebook group at `facebook.com/groups/198849596348`) is already tracked in **#1159** from the prior session. Profile completeness (founded 1995, schedule, hash cash, description, website) and event extraction (all 12 events from the HashNYC HTML scraper source) are verified accurate.

When the admin-tooling failure is fixed, an admin can mark GGFM complete on my behalf with:
- Findings filed: 0
- Summary: "No new findings — #1159 covers the only remaining profile gap; HashNYC HTML scraper extraction verified accurate against source."

## See also

- #1207 — original strict report for GLH3 (same failure mode this session)
- #1160 — earlier *misattribution* mode (different bug; in that case submit went to a different kennel rather than failing silently)

johnrclem:
Closes #1207
Closes #1216
Closes #1175

## Summary

Fixes the admin audit "Mark deep dive complete" dialog which intermittently rendered a *"Server Components render"* error and left submit as a silent no-op for GLH3 (#1207) and GGFM (#1216). Also addresses the residual a11y asks from #1175 not already covered by PR #1203's snapshot-token guard.

## Root cause

PR #1203 introduced the `getDeepDiveQueueToken` server action and the dialog called it from a `useEffect` on every open. Under intermittent Prisma `TooManyConnections` pressure (visible in production logs around the same window) that action threw. Next.js sanitized the message to *"An error occurred in the Server Components render…"* and the dialog wrote that text into its error state. Submit button was `disabled={!queueToken}`, so the click looked like a silent no-op.

`AUDIT_QUEUE_TOKEN_SECRET` is set in Vercel — the missing-secret theory was ruled out.

## Fix

- **Mint queue tokens at page render** (`src/lib/queue-snapshot-token.ts`) — one batch pass over the already-fetched queue, sharing the snapshot ID + expiration. Returns `{}` on signing failure so the page still renders and the dialog falls back to its async path.
- **Pass tokens through props** (`src/app/admin/audit/page.tsx` → `src/components/admin/AuditDashboard.tsx`) — eliminates the on-open server action call from the happy path entirely.
- **Freshness gate** — Codex flagged that a tab idle past the 10-min TTL would submit a stale token and the user couldn't recover via reopen. Tokens with ≤30s of TTL fall through to the async fetch.
- **In-place refresh on `invalidToken`** — same retry path the `queueChanged` branch already uses, so a long-idle dashboard recovers without a page reload.
- **Sentry capture in the fallback action** — production sanitization swallows the digest from the user; Sentry preserves the actual error.
- **`data-kennel-code` attributes** on the `Select` trigger, items, and queue rows — closes #1175 ask 6 (stable read path for accessibility-tree agents).

## Why this approach

`getDeepDiveQueueToken` already duplicated work the parent page did 1s earlier. Moving the mint to the page render uses data we already have, drops three Prisma calls per dialog open, and routes around the `TooManyConnections` failure surface entirely. The async path stays as the fallback for stale-prop and edited-queue cases, and the existing snapshot-binding (`computeQueueSnapshotId`) keeps #1160's misattribution defense intact — the server still recomputes from live state at submit time.

#1175 asks 6 (stable a11y attribute) and 8 (server-side guard) are now both addressed; ask 7 (virtualization) is moot since the row-click pattern landed in #1203 doesn't virtualize.

## Test plan

- [x] `npm test` — 6001 passed (1 new test added; Codex regression test for token expiry + freshness gate)
- [x] `npm run lint` — clean (0 errors)
- [x] `/codex:adversarial-review` flagged the TTL/recovery gap; addressed in this PR
- [x] `/simplify` 3-agent pass; type/comment polish applied
- [ ] Manual: open `/admin/audit`, click Mark complete on a queue kennel, verify submit succeeds and queue updates
- [ ] Manual: hard-reload after submit, verify `last dived` shows today and coverage tick increments
- [ ] Manual: leave dashboard idle >10 min, click Mark complete, verify dialog re-fetches token instead of failing
- [ ] Post-merge: mark GLH3 (#1207) and GGFM (#1216) complete on production using the summaries from those issue bodies

### DIFF
--- a/src/app/admin/audit/actions.ts
+++ b/src/app/admin/audit/actions.ts
@@ -1,5 +1,6 @@
 "use server";
 
+import * as Sentry from "@sentry/nextjs";
 import { Prisma, AuditStream, AuditIssueEventType } from "@/generated/prisma/client";
 import { prisma } from "@/lib/db";
 import { getAdminUser } from "@/lib/auth";
@@ -505,16 +506,24 @@ export async function getDeepDiveCoverage(): Promise<DeepDiveCoverage> {
 export async function getDeepDiveQueueToken(
   kennelCode: string,
 ): Promise<{ token: string; expiresAt: number } | null> {
-  await requireAdmin();
-  if (!kennelCode) return null;
+  try {
+    await requireAdmin();
+    if (!kennelCode) return null;
 
-  const kennelCodes = await getDeepDiveQueueKennelCodes();
-  if (!kennelCodes.includes(kennelCode)) return null;
+    const kennelCodes = await getDeepDiveQueueKennelCodes();
+    if (!kennelCodes.includes(kennelCode)) return null;
 
-  const expiresAt = computeQueueTokenExpiresAt();
-  const queueSnapshotId = computeQueueSnapshotId(kennelCodes);
-  const token = signQueueToken({ kennelCode, queueSnapshotId, expiresAt });
-  return { token, expiresAt };
+    const expiresAt = computeQueueTokenExpiresAt();
+    const queueSnapshotId = computeQueueSnapshotId(kennelCodes);
+    const token = signQueueToken({ kennelCode, queueSnapshotId, expiresAt });
+    return { token, expiresAt };
+  } catch (err) {
+    // Production-sanitized "Server Components render" error hides the
+    // root cause from the dialog. Capture before rethrowing so the
+    // actual digest stays diagnosable in Sentry (issues #1207 / #1216).
+    Sentry.captureException(err, { extra: { kennelCode } });
+    throw err;
+  }
 }
 
 /** Result shape for `recordDeepDive`. Discriminated union on `ok`
--- a/src/app/admin/audit/page.tsx
+++ b/src/app/admin/audit/page.tsx
@@ -15,6 +15,7 @@ import {
 import { KNOWN_AUDIT_RULES } from "@/pipeline/audit-checks";
 import { AuditDashboard } from "@/components/admin/AuditDashboard";
 import { buildHarelinePrompt } from "@/lib/admin/hareline-prompt";
+import { mintQueueTokens } from "@/lib/queue-snapshot-token";
 
 /** Build the daily hareline audit prompt at request time so the curated
  *  sections (recently-fixed, focus areas) reflect live data. Returns null on
@@ -67,6 +68,16 @@ export default async function AuditPage() {
     getRecentOpenIssues().catch(() => []),
   ]);
 
+  // Pre-mint a queue token per candidate at page render so the dialog
+  // doesn't have to round-trip a server action on open. Eliminates the
+  // "Server Components render" error path from #1207 / #1216 caused by
+  // intermittent failures in the on-open token mint. `mintQueueTokens`
+  // returns `{}` on failure (e.g. missing secret), in which case the
+  // dialog falls back to its async `getDeepDiveQueueToken` path.
+  const deepDiveTokens = mintQueueTokens(
+    deepDiveQueueResult.map((k) => k.kennelCode),
+  );
+
   return (
     <AuditDashboard
       trends={trendsResult}
@@ -77,6 +88,7 @@ export default async function AuditPage() {
       knownRules={[...KNOWN_AUDIT_RULES]}
       deepDiveQueue={deepDiveQueueResult}
       deepDiveCoverage={deepDiveCoverageResult}
+      deepDiveTokens={deepDiveTokens}
       harelinePrompt={harelinePrompt}
       streamTrends={streamTrendsResult}
       streamOpenCounts={streamOpenCountsResult}
--- a/src/components/admin/AuditDashboard.tsx
+++ b/src/components/admin/AuditDashboard.tsx
@@ -83,6 +83,18 @@ interface KennelOption {
   shortName: string;
 }
 
+import type {
+  MintedQueueToken,
+  MintedQueueTokens,
+} from "@/lib/queue-snapshot-token";
+
+/** Pre-minted queue tokens keyed by kennelCode. Empty when the
+ *  page-render mint failed (e.g. missing secret) — the dialog falls
+ *  back to the async `getDeepDiveQueueToken` path. Re-exports the
+ *  lib type so the `page → dashboard → dialog` prop chain stays in
+ *  one shape; rename to `DeepDiveTokens` for context at call sites. */
+export type DeepDiveTokens = MintedQueueTokens;
+
 interface Props {
   trends: TrendPoint[];
   topOffenders: TopOffender[];
@@ -92,6 +104,7 @@ interface Props {
   knownRules: string[];
   deepDiveQueue: DeepDiveCandidate[];
   deepDiveCoverage: DeepDiveCoverage;
+  deepDiveTokens: DeepDiveTokens;
   harelinePrompt: string | null;
   streamTrends: StreamTrendPoint[];
   streamOpenCounts: StreamOpenCounts[];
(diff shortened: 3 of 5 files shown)


johnrclem:
@claude @codex review

PR #1341, opened 2026-05-10
chore(scripts): one-shot RawEvent fingerprint dedup (pre #1286 migration)

johnrclem:
## Summary

One-shot cleanup script that collapses existing duplicate `(sourceId, fingerprint)` groups in `RawEvent` ahead of issue #1286's `@@unique([sourceId, fingerprint])` migration. Without this cleanup the migration would fail on the existing dupe rows; this PR ships the script first so we can run it against prod and verify `COUNT(*) = 0` **before** opening the migration PR.

This PR is review-only — **no DB writes happen on merge.** I'll run \`npx tsx scripts/dedup-rawevent-fingerprint.ts --apply\` against prod after merge, attach the audit log, then open the migration PR.

## Pre-flight diagnostics (2026-05-10, prod)

| Metric | Count |
|---|---|
| Duplicate \`(sourceId, fingerprint)\` groups | **766** |
| Doomed rows (count − 1 per group) | **7,647** |
| Groups with NO linkage (orphan dupes) | 468 |
| Groups with linkage | 298 |
| Multi-link groups (≥2 rows linked) | 136 |
| Multi-link groups with split canonical Events | **0** |
| Max distinct \`eventId\`s in any one group | 1 |

The "split canonical Event = 0" outcome simplifies survivor selection: every linked group's linked rows point to the *same* canonical Event, so picking any linked-row survivor naturally inherits the linkage. The "transfer eventId from doomed → survivor" path my plan called for is unneeded.

Top sources by doomed row count:
- Whoreman H3 Calendar (1,495), Boston Hash Calendar (1,438), Princeton NJ Hash Calendar (1,309), HashNYC Website (601) — high-multiplicity dupes (~50 dupes/group on average), suggest a long-running issue where the per-event \`findFirst\` dedup was bypassed.
- Charm City iCal (81/81), Bangkok Full Moon (65/65), Hudson Valley Meetup (10/30) — exactly 2 dupes/group, the classic concurrent-scrape race.

Whatever the historical root causes, the new \`@@unique([sourceId, fingerprint])\` constraint (issue #1286) closes the door going forward.

## Survivor selection

\`\`\`ts
// pickSurvivor (pure function, 5 unit tests)
// 1. If any sibling has \`eventId IS NOT NULL\`, pick the linked row with
//    most-recent \`scrapedAt\` (naturally inherits canonical Event linkage).
// 2. Else (no linkage), pick the OLDEST sibling by \`scrapedAt\` —
//    preserves the original observation timestamp.
[...]

PR #1280, opened 2026-05-07
perf(merge): batch RawEvent dedup prefetch to fix N+1 in scrape pipeline

johnrclem:
## Summary

Fixes the Sentry [JAVASCRIPT-NEXTJS-3](https://hashtracks.sentry.io/issues/7375103156/) N+1: `POST /api/cron/scrape/[sourceId]` was issuing one `prisma.rawEvent.findFirst({ fingerprint, sourceId })` per scraped event (376 occurrences since 2026-03-30, ~16 repeated DB spans per request, 5.7s sample transaction).

- Replace the per-event `findFirst` with a single `rawEvent.findMany({ fingerprint: { in: [...] } })` prefetch before the merge loop, keyed into a `Map<fingerprint, ExistingRawEventEntry>` on `MergeContext`.
- Fan the prefetch out via `Promise.all` alongside `source.findUnique` and `sourceKennel.findMany` — saves a round-trip on every scrape, and chunks at 1000 fingerprints to keep planner cost linear on historical replays (SDH3 ~7649 events).
- Mirror new RawEvents into the in-memory map immediately after `upsertCanonicalEvent` marks them processed, so an in-batch duplicate fingerprint still takes the existing-Event branch even when a later side-effect throws.

Drops dedup-phase DB round-trips from O(events) to O(1).

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npm test` — full 6007-test suite green; merge.test.ts covers 287 cases including two new regressions (one asserting exactly one `findMany({ fingerprint: { in } })` per batch, one asserting the in-memory write-back handles same-fingerprint dupes within a batch)
- [x] `npm run lint` clean for the changed files
- [ ] Production verify: after deploy, confirm the next QStash-driven scrape's Sentry trace collapses the repeated `prisma:client:db_query` group from N to 1, and that the issue's occurrences counter stops incrementing within ~24 hrs.

## Out-of-scope follow-ups (intentionally not in this PR)

- Add `@@unique([sourceId, fingerprint])` on `RawEvent` to close the cross-worker race window the prefetch widens (existed before, just smaller). Needs a migration that may have to dedupe historical rows first.
- Batch `event.findMany({ kennelId, date })` sameDayEvents in `upsertCanonicalEvent` (line 921) and the fuzzy `event.findMany({ kennelId, date in window })` (line 865) for new-event-heavy scrapes.
[...]

PR #1355, opened 2026-05-10
perf(merge): @@unique([sourceId, fingerprint]) on RawEvent + P2002 fallthrough

### LINKED ISSUE #1286: Add @@unique([sourceId, fingerprint]) on RawEvent to close concurrent-scrape race
## Background

PR #1280 fixed the Sentry N+1 ([JAVASCRIPT-NEXTJS-3](https://hashtracks.sentry.io/issues/7375103156/)) by replacing per-event `prisma.rawEvent.findFirst` dedup with a single batch prefetch. This widened the cross-worker race window between "check fingerprint exists" and "create RawEvent" — flagged by Codex's adversarial review on the PR.

The race existed before too (just per-event-sized), but `RawEvent` only has `@@index([fingerprint])`, no uniqueness guarantee. Two QStash workers scraping the same source concurrently (e.g. on retry after a 300s timeout) can both miss the prefetch and both insert the same `(sourceId, fingerprint)` row, leading to:
- duplicate `RawEvent` rows
- duplicate canonical-event processing for the same scraped row

## Proposal

Add a unique constraint on `(sourceId, fingerprint)`:

```prisma
model RawEvent {
  // …
  @@unique([sourceId, fingerprint])
}
```

Then `processNewRawEvent`'s `prisma.rawEvent.create` will throw `P2002` on a concurrent dupe, and we can catch + read-through to the existing row (effectively the per-event `findFirst` for race-loser worker only).

## Risks / prereqs

- **Existing dupes**: there may already be `(sourceId, fingerprint)` duplicates from past races. The migration must dedupe them first or it'll fail. Recommended approach:
  1. Author a migration that runs a `DELETE FROM "RawEvent" a USING "RawEvent" b WHERE …` keeping the row with the linked canonical Event (or oldest) and removing siblings.
  2. Add the unique constraint in the same migration.
  3. Verify against the local Postgres DB first per `.claude/rules/local-dev-db.md`.
- **Failure handling**: `processNewRawEvent` needs a P2002 catch block that re-fetches the existing row and re-routes to the existing-Event branch, mirroring `handleDuplicateFingerprint`.

## Files to touch

- `prisma/schema.prisma` — add `@@unique([sourceId, fingerprint])` on `RawEvent`
- `prisma/migrations/<ts>_rawevent_sourceid_fingerprint_unique/migration.sql` — dedupe + add constraint
- `src/pipeline/merge.ts` — wrap `prisma.rawEvent.create` in try/catch for `P2002`, fall back to fetching the existing row and routing through the existing-Event branch
- `src/pipeline/merge.test.ts` — add a regression test simulating the P2002 race

Linked: PR #1280 (deferred from)

johnrclem:
## Summary

Closes the cross-worker race window the dedup prefetch in PR #1280 widened. Two concurrent QStash workers scraping the same source can both miss the per-batch fingerprint prefetch and both reach `prisma.rawEvent.create`. Without a uniqueness guarantee on `(sourceId, fingerprint)`, both inserts succeeded and produced duplicate canonical-event work — the same root cause that produced the 7,697 historical duplicates cleaned up in PR #1341.

This PR is the third and final step:
- **PR #1280** — N+1 fix (per-event `findFirst` → batch prefetch)
- **PR #1341** — one-shot cleanup script + ran `--apply` against prod (7,697 rows deleted)
- **This PR** — schema constraint (closes the race forever) + runtime P2002 fall-through (race-loser worker silently routes through the duplicate path)

## Schema change

```prisma
model RawEvent {
  // ...
  @@unique([sourceId, fingerprint])
  @@index([sourceId, scrapedAt])
}
```

The standalone `@@index([fingerprint])` is dropped — every fingerprint query in the codebase also filters by sourceId (the dedup prefetch in `processRawEvents → prefetchExistingByFingerprint`), so the new composite unique index covers them with a leftmost-prefix match. Keeping the standalone would just be redundant write amplification on every RawEvent insert.

## Migration

`prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql`:

1. **Defensive WITH-DELETE** — same survivor algorithm as the script (linked rows beat unlinked, most-recent scrapedAt within linked, oldest within unlinked, `id` ASC as the deterministic final tiebreaker). Atomic with the index creation. PR #1341's `--apply` already brought prod to 0 dupes; this defensive sweep catches any race-window dupes that landed between the script and this migration's deploy.
2. **Drop `RawEvent_fingerprint_idx`** (the standalone index that the schema change retires).
3. **Create `RawEvent_sourceId_fingerprint_key`** (the new unique constraint).

All three statements run inside Prisma's per-migration transaction, so the schema change is atomic with the cleanup.

## Runtime P2002 handler

Wraps `prisma.rawEvent.create` in `processNewRawEvent` with try/catch. New module-level helper `isRawEventFingerprintP2002` narrows P2002s to the `(sourceId, fingerprint)` constraint specifically — other P2002s (future schema additions) bubble unchanged.

On conflict:
1. Re-fetch the winning row via `prisma.rawEvent.findUnique({ where: { sourceId_fingerprint } })`.
2. Mirror it into `ctx.existingByFingerprint`.
3. Route through `handleDuplicateFingerprint`. Net effect: the racing-loser worker counts the event as `skipped` instead of bubbling as `eventErrors`.

The handler distinguishes three duplicate-handler return values (`string | null | false`) — the `false` branch (winner is an unprocessed orphan asking for re-processing) leaves the orphan for the racing worker to resolve and counts the event as `skipped` from our perspective.

## Tests (5 new, 293 total in `merge.test.ts`)

- Routes through duplicate path when winner is processed + linked.
- Routes through duplicate path when winner is unprocessed orphan.
- Re-throws P2002 targeting different columns (defensive).
- Re-throws non-P2002 errors (e.g. P2003 FK violation).
- Multi-event batch: P2002 on event 1, normal create on event 2 — counts 1 created + 1 skipped, no double-counting.

## Test plan

- [x] `npx tsc --noEmit` clean (only pre-existing `suncalc` errors)
- [x] `npm test src/pipeline/merge.test.ts` — 293/293 passing
- [x] `npm run lint` clean for changed files
- [x] PR #1341's `--apply` ran against prod; verified `COUNT(*) = 0` for `(sourceId, fingerprint)` duplicate groups before this PR was opened
- [x] `/simplify` review — extracted `isRawEventFingerprintP2002` helper, trimmed P2002 handler comments, dropped unnecessary `as string[]` casts, removed redundant test mock setup
- [ ] Post-merge: verify the constraint exists in prod via `SELECT indexname FROM pg_indexes WHERE tablename = 'RawEvent';`
- [ ] Post-merge: tail Sentry for any P2002 events on `RawEvent.create` over the next 24h — expected zero, OR a small handful that route through the duplicate path silently (no `eventErrors` increment, no Sentry capture).

Closes #1286

### DIFF
--- a/prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql
+++ b/prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql
@@ -17,6 +17,16 @@
 --      script's localeCompare-based JS tiebreak so a partial
 --      script run + this migration's defensive sweep agree on the
 --      survivor).
+--
+-- Lock the table for the duration of the migration so concurrent QStash
+-- workers can't INSERT a fresh duplicate between the DELETE and the
+-- CREATE UNIQUE INDEX (which would fail with a unique-violation and roll
+-- back the whole migration). ACCESS EXCLUSIVE is what CREATE INDEX takes
+-- anyway; acquiring it up-front folds a few hundred ms of would-be
+-- writer-blocking into a single contiguous window. Released on
+-- COMMIT/ROLLBACK of the implicit migration transaction.
+LOCK TABLE "RawEvent" IN ACCESS EXCLUSIVE MODE;
+
 WITH ranked AS (
   SELECT id,
     ROW_NUMBER() OVER (
--- a/prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql
+++ b/prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql
@@ -1,0 +1,43 @@
+-- Close the cross-worker race window in merge.ts by adding
+-- @@unique([sourceId, fingerprint]) on RawEvent (issue #1286).
+--
+-- Without this constraint, two concurrent QStash workers scraping the
+-- same source can both miss the dedup prefetch (PR #1280) and both
+-- insert the same row. Pre-deploy cleanup collapsed 7,697 historical
+-- dupes (PR #1341 / scripts/dedup-rawevent-fingerprint.ts). The
+-- defensive WITH-DELETE below is a no-op when there are no remaining
+-- duplicate groups; it catches any race-window dupes that landed
+-- between the pre-deploy script run and this migration's deploy.
+--
+-- Survivor selection mirrors the script's pickSurvivor algorithm:
+--   1. Linked rows beat unlinked (eventId IS NOT NULL ranks first).
+--   2. Among linked siblings, most-recent scrapedAt wins.
+--   3. Among unlinked siblings, OLDEST scrapedAt wins.
+--   4. id ASC is the deterministic final tiebreaker (matches the
+--      script's localeCompare-based JS tiebreak so a partial
+--      script run + this migration's defensive sweep agree on the
+--      survivor).
+WITH ranked AS (
+  SELECT id,
+    ROW_NUMBER() OVER (
+      PARTITION BY "sourceId", "fingerprint"
+      ORDER BY
+        CASE WHEN "eventId" IS NOT NULL THEN 0 ELSE 1 END ASC,
+        CASE WHEN "eventId" IS NOT NULL THEN "scrapedAt" END DESC NULLS LAST,
+        CASE WHEN "eventId" IS NOT NULL THEN NULL ELSE "scrapedAt" END ASC NULLS LAST,
+        id ASC
+    ) AS rn
+  FROM "RawEvent"
+)
+DELETE FROM "RawEvent" WHERE id IN (SELECT id FROM ranked WHERE rn > 1);
+
+-- Drop the standalone fingerprint index. Every RawEvent fingerprint
+-- query in the codebase also filters by sourceId (the dedup prefetch in
+-- merge.ts at processRawEvents → prefetchExistingByFingerprint), so the
+-- new composite unique index below covers them with a leftmost-prefix
+-- match. Keeping the standalone index would just be redundant write
+-- amplification on every RawEvent insert.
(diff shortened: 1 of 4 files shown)


johnrclem:
@claude @codex review

johnrclem:
Pushed `237d0e04` addressing the review feedback:

**1. Migration race (Codex P1)** — added `LOCK TABLE "RawEvent" IN ACCESS EXCLUSIVE MODE;` at the top of the migration. CREATE INDEX would take this lock anyway, so the change just folds the few-hundred-ms writer-blocking window into one contiguous span at the start, preventing concurrent QStash workers from inserting between the DELETE and CREATE INDEX.

**2. Helper precision (Gemini Medium + @claude review)** — added `target.length === 2` guard to `isRawEventFingerprintP2002` so a future composite unique constraint that happens to include `sourceId` and `fingerprint` plus other columns doesn't silently match.

**3. Orphan stuck-skip (Codex P1)** — fixed the regression. The catch-block now ADOPTS the unprocessed orphan and runs the rest of `processNewRawEvent` against `winner.id` instead of skipping. Pre-#1286 a crashed-mid-processing worker would leave a tombstone but the event got re-created; the unique constraint blocked that re-creation, dropping the event permanently. The 468 unlinked-survivor orphans left by the cleanup script will now be processed correctly on their next scrape.

**Test changes:**
- The existing orphan-skip test is now the orphan-adopt test — asserts `created: 1, skipped: 0` and verifies `mockRawEventUpdate` was called against the orphan's `id`.
- New test for the `findUnique → null` TOCTOU branch (winner deleted between P2002 and re-fetch — surfaces as `eventErrors`, not silently dropped).

294/294 tests pass; tsc + lint clean.

PR #1356, opened 2026-05-10
refactor: extract isUniqueConstraintViolation + buildPrismaUniqueViolation

johnrclem:
## Summary

Deferred reuse cleanup from #1355 (the issue #1286 race-window handler). The codebase had ~10 inline copies of the same Prisma P2002 detection shape (`err instanceof Prisma.PrismaClientKnownRequestError && err.code === "P2002"` ± `target` array narrowing). Consolidating into one shared helper:
- Eliminates the recurring pattern (10 sites → 1 helper).
- Makes the AND-of-tuple vs no-arg semantics explicit at one documented site.
- Gives tests a single P2002 builder factory.

> **Stacked on #1355.** Until that PR merges, this PR's diff includes the P2002 try/catch in `src/pipeline/merge.ts` from #1355. Once #1355 merges, the diff shrinks to just the refactor (the migrated `merge.ts` site stays, but the surrounding handler becomes part of `main`).

## New helpers

**`src/lib/prisma-errors.ts`** — `isUniqueConstraintViolation(err, requiredTargetColumns?)` type guard. Pass a column tuple to narrow to a specific composite constraint (AND semantics, since unique constraints target a fixed tuple). Use the no-arg form when several single-column unique constraints could plausibly fire at one call site.

**`src/test/factories.ts`** — `buildPrismaUniqueViolation(target)` factory, replacing repeated `new Prisma.PrismaClientKnownRequestError(...)` construction across 5 test files.

## Migrated call sites (production)

| File | Form | Notes |
|---|---|---|
| `src/pipeline/merge.ts` | `["sourceId", "fingerprint"]` | drops local `isRawEventFingerprintP2002` from #1355 |
| `src/app/admin/shared/source-errors.ts` | `["name", "type"]` | replaces inline `target.includes` chain |
| `src/lib/auth.ts` | no-arg | clerkId or email could fire — single-column constraints, AND mode would never match |
| `src/lib/kennel-utils.ts` | no-arg | |
| `src/app/logbook/actions.ts` | no-arg | 2 sites |
| `src/app/admin/audit/actions.ts` | no-arg | |
| `src/app/admin/research/actions.ts` | no-arg | (not on the original list — straggler caught) |
| `src/app/travel/actions.ts` | no-arg | 3 sites |
| `src/pipeline/kennel-discovery.ts` | no-arg | |

## Migrated test files

`merge.test.ts`, `auth.test.ts`, `logbook/actions.test.ts`, `travel/actions.test.ts` (3 sites), `audit/actions.test.ts`. The audit test mocks `@/generated/prisma/client` with its own `PrismaClientKnownRequestError` class; the factory works through that mock because Vitest module-caches resolve both `factories.ts` and the helper to the same mocked class — verified by running the test.

## Test plan

- [x] `npx tsc --noEmit` clean (no errors from changed files; pre-existing `suncalc` resolution errors went away after `npm install`)
- [x] `npm test` — 6,306/6,306 passing
- [x] `/simplify` review (3 parallel agents) — applied 2 small findings: trimmed misleading `["clerkId"]` JSDoc example in favor of an AND-semantics note, shortened the `auth.ts` comment from 4 lines to 1.

## Net diff

15 files modified, 1 new file. **+85 / -81 lines** — net -19 despite adding two helpers, because each call site shrinks from ~3-5 lines to 1.

### DIFF
--- a/prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql
+++ b/prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql
@@ -17,6 +17,16 @@
 --      script's localeCompare-based JS tiebreak so a partial
 --      script run + this migration's defensive sweep agree on the
 --      survivor).
+--
+-- Lock the table for the duration of the migration so concurrent QStash
+-- workers can't INSERT a fresh duplicate between the DELETE and the
+-- CREATE UNIQUE INDEX (which would fail with a unique-violation and roll
+-- back the whole migration). ACCESS EXCLUSIVE is what CREATE INDEX takes
+-- anyway; acquiring it up-front folds a few hundred ms of would-be
+-- writer-blocking into a single contiguous window. Released on
+-- COMMIT/ROLLBACK of the implicit migration transaction.
+LOCK TABLE "RawEvent" IN ACCESS EXCLUSIVE MODE;
+
 WITH ranked AS (
   SELECT id,
     ROW_NUMBER() OVER (
--- a/prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql
+++ b/prisma/migrations/20260510141000_rawevent_sourceid_fingerprint_unique/migration.sql
@@ -1,0 +1,43 @@
+-- Close the cross-worker race window in merge.ts by adding
+-- @@unique([sourceId, fingerprint]) on RawEvent (issue #1286).
+--
+-- Without this constraint, two concurrent QStash workers scraping the
+-- same source can both miss the dedup prefetch (PR #1280) and both
+-- insert the same row. Pre-deploy cleanup collapsed 7,697 historical
+-- dupes (PR #1341 / scripts/dedup-rawevent-fingerprint.ts). The
+-- defensive WITH-DELETE below is a no-op when there are no remaining
+-- duplicate groups; it catches any race-window dupes that landed
+-- between the pre-deploy script run and this migration's deploy.
+--
+-- Survivor selection mirrors the script's pickSurvivor algorithm:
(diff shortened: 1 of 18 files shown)


johnrclem:
@claude @codex review

johnrclem:
Pushed `f5c6133a` (merge commit from `claude/perf-merge-rawevent-unique`) addressing review feedback:

**1. Set-equality match (Gemini Medium + @claude review)** — `isUniqueConstraintViolation` now requires `target.length === requiredTargetColumns.length` in addition to `every-includes`, so a future composite constraint that happens to include the caller's columns (plus others) doesn't silently fall through. Matches the documented "fixed tuple" semantics.

**2. @codex P1 — generated client missing `sourceId_fingerprint`** — false positive. `src/generated/prisma/` is gitignored (verified via `git ls-files`); Vercel runs `prisma generate` via `postinstall` (`package.json:"postinstall": "prisma generate"`), so the generated client is always rebuilt from the schema before runtime. Locally, my tests pass because `npx prisma generate` runs as part of the dev workflow.

**3. @codex P1 — orphan stuck-skip** — fixed via the merged-in commit from PR #1355. The catch-block now adopts unprocessed orphans and runs the rest of `processNewRawEvent` against `winner.id`. Test now asserts `created: 1, skipped: 0` for the orphan-adopt case + new test for the `findUnique → null` TOCTOU branch.

The merge commit brings PR #1355's fixes into this branch (migration LOCK, length guard on the local helper that this PR removes, orphan-adopt). Once #1355 merges first, this PR's diff vs main will naturally show only the refactor — the fixes propagate cleanly through merge order.

294/294 merge tests + 6,307 full-suite tests pass; tsc + lint clean.

Skipping CodeRabbit's rate-limited review and Gemini's overlapping length-equality suggestion (same fix as #1).

PR #1359, opened 2026-05-10
perf(merge): batch sameDayEvents, fuzzy ±48h, lastEventDate per kennel

### LINKED ISSUE #1287: Batch remaining per-event Prisma queries in merge pipeline (sameDayEvents, fuzzy, lastEventDate)
## Background

PR #1280 collapsed the dedup-phase N+1 in `processRawEvents` (`prisma.rawEvent.findFirst` → batch prefetch). Three other per-event Prisma calls remain inside the loop. They only fire for *new* events that pass dedup, so a steady-state scrape (mostly duplicates) is unaffected — but they're the next set of N+1s once Sentry's spotlight moves on.

## Targets

All inside `src/pipeline/merge.ts`. Line refs are post-merge of PR #1280.

| # | Call site | What | Frequency |
|---|---|---|---|
| 1 | `upsertCanonicalEvent` line ~921 | `prisma.event.findMany({ where: { kennelId, date }, … })` — same-day-events lookup | Once per new event |
| 2 | `findFuzzyDuplicateInWindow` line ~865 | `prisma.event.findMany({ where: { kennelId, date: { gte, lte, not } }, … })` — ±48h fuzzy candidates | Once per new event |
| 3 | `processNewRawEvent` line ~1416 | `prisma.\$executeRaw UPDATE \"Kennel\" SET \"lastEventDate\"` | Once per new event |

## Proposed approach

**#1 + #2** — both are keyed by `(kennelId, date or date-window)`. Group events by `kennelId` upfront, then issue one `findMany` per kennel covering the union date range, materialize into a per-kennel sorted array, and split per-event during the loop.

**#3** — kennel `lastEventDate` is a max aggregate. After the loop, run one `updateMany` per source covering all touched kennels:
```ts
const maxDateByKennel = new Map<string, Date>();
// ... populate during loop ...
await prisma.\$transaction(
  [...maxDateByKennel].map(([kennelId, maxDate]) =>
    prisma.kennel.updateMany({
      where: { id: kennelId, OR: [{ lastEventDate: null }, { lastEventDate: { lt: maxDate } }] },
      data: { lastEventDate: maxDate, updatedAt: new Date() },
    }),
  ),
);
```

## Verification

- Add a regression test asserting at most 1 \`event.findMany\` per distinct kennelId in a multi-event-multi-kennel batch.
- Sentry trace check on a chatty source (e.g. SDH3 forced re-scrape) — span tree should show one event.findMany per kennel, not per event.

Linked: PR #1280 (deferred from)

johnrclem:
## Summary

Closes the last three per-new-event Prisma reads in `processRawEvents`, deferred from PR #1280's N+1 fix:

- `event.findMany({ kennelId, date: eventDate })` (sameDayEvents) — once per new event in `upsertCanonicalEvent`.
- `event.findMany({ kennelId, date: { gte, lte, not }, parentEventId: null, isSeriesParent: false })` (fuzzy ±48h candidates) — once per new event in `findFuzzyDuplicateInWindow`.
- `$executeRaw UPDATE "Kennel" SET "lastEventDate"` — once per new event in `processNewRawEvent`.

These only fire for events that pass the dedup prefetch (i.e. genuinely new), so steady-state mostly-duplicate scrapes don't see them — but backfills and first scrapes do. The N+1 was the next bottleneck after PR #1280 closed the Sentry-flagged one.

## Approach

**Per-kennel cache, lazy-populated.** `ensureKennelEventCache(kennelId, ctx)` runs ONE `event.findMany` per kennel touched in a batch, populating:
- `ctx.sameDayCache: Map<kennelId, EventRow[]>` — flat list per kennel; `getSameDayEvents` filters by exact date on access.
- `ctx.fuzzyPoolByKennel: Map<kennelId, EventRow[]>` — pre-filtered to parents/non-series, pre-sorted by trustLevel desc.

**Window strategy adapts to the fuzzy flag:**
- `MERGE_FUZZY_DEDUP=true` (off in prod): widens to `date BETWEEN min(batchDates) - 48h AND max(batchDates) + 48h`.
- Default: `date: { in: [...batchDates] }` — narrow, avoids the SDH3-backfill pathology where one scrape would pull 7K+ historical rows because `batchDates` spans a year.

**`rememberCreatedEvent` mirrors fresh creates** into both caches via an O(N) insertion-sort splice (not `Array.sort`), keeping a batch of M new-event creations linear in fuzzy-pool size, not M·N·logN.

**`invalidateKennelEventCache` runs only inside the trust-gated UPDATE branch** where `date` actually changes in DB. Lower-trust enrichment paths don't move the row, so the cache stays valid and we skip a redundant refetch.

**Per-event `$executeRaw` replaced by `ctx.kennelMaxDates: Map<kennelId, Date>`,** flushed as one `Promise.all` of `kennel.updateMany` after the loop. Preserves the "only update if newer" semantic via the `OR: [{ lastEventDate: null }, { lastEventDate: { lt: maxDate } }]` guard.

## Query count change

For a batch of M new events on K distinct kennels:

| Read | Pre-#1287 | Post-#1287 |
|---|---|---|
| sameDayEvents | M `event.findMany` | K `event.findMany` (with sameDayCache + fuzzy pool) |
| Fuzzy ±48h candidates | M `event.findMany` | 0 (in-memory from sameDay prefetch) |
| Kennel.lastEventDate | M `$executeRaw` | K `kennel.updateMany` (parallel) |
| **Total per-event reads** | **2M + M writes** | **K reads + K writes** |

For the typical scrape (M=20–50 events on K=1 kennel): 60–150 round-trips → 2 round-trips. For SDH3-style backfill (M=7,649 on K=10): 22,947 round-trips → 20 round-trips.

## Tests

- 4 new regressions in `merge.test.ts`:
  - "issues at most one event.findMany per distinct kennel across a multi-event batch (fuzzy off)" — verifies `date: { in: batchDates }` narrow shape.
  - "widens the cache prefetch to ±48h around batch dates when MERGE_FUZZY_DEDUP is on" — verifies `date: { gte, lte }` wide shape.
  - "issues at most one kennel.updateMany per distinct kennel" — verifies the batched flush + max-date guard.
  - "does not call kennel.updateMany when no new events are processed (all duplicates)" — verifies the dedup-only path skips the flush.
- Two existing fuzzy tests updated to match the new query order (single cache prefetch + cross-window old-bucket refetch instead of three separate findMany calls).
- Added `vi.mocked(prisma.event.findMany).mockReset()` in `beforeEach` to prevent leftover `mockResolvedValueOnce` queues from prior tests being consumed by the new prefetch — same hygiene fix PR #1280 made for `rawEvent.findMany`.

## Test plan

- [x] `npm test src/pipeline/merge.test.ts` — 298/298 passing (4 new + 2 updated fuzzy + 1 new for cross-source race-window cleanup)
- [x] `npm test` (full suite) — 6,311 passing, 7 skipped, 25 todo
- [x] `/simplify` review applied 4 fixes (window gating, insertion sort, trust-gated cache invalidation, comment trim)
- [ ] Production verify post-deploy: confirm a chatty source's Sentry trace shows 1 `event.findMany` per kennel and 1 `kennel.updateMany` per kennel, not per new event

Closes #1287

### DIFF
--- a/src/pipeline/merge.test.ts
+++ b/src/pipeline/merge.test.ts
@@ -3124,7 +3124,10 @@ describe("fuzzy ±48h cross-source dedup (#990)", () => {
     // catches this via a RawEvent join even when title/time/location all
     // pass.
     mockRawEventFind.mockResolvedValueOnce(null);
-    mockEventFindMany.mockResolvedValueOnce([] as never); // same-day empty
+    // Cache prefetch returns the candidate; getSameDayEvents filters it
+    // out of the incoming bucket (different date) and the fuzzy probe
+    // picks it up from the in-memory pool (#1287). The wide ±48h prefetch
+    // window (MERGE_FUZZY_DEDUP=true in beforeEach) keeps it visible.
     mockEventFindMany.mockResolvedValueOnce([
       existingFuzzyRow({ id: "evt_same_source_yesterday" }),
     ] as never);
@@ -3151,7 +3154,10 @@ describe("fuzzy ±48h cross-source dedup (#990)", () => {
 
   it("does NOT merge when runNumber differs (double-header on consecutive days)", async () => {
     mockRawEventFind.mockResolvedValueOnce(null);
-    mockEventFindMany.mockResolvedValueOnce([] as never); // same-day empty
+    // Cache prefetch returns the candidate; getSameDayEvents filters it
+    // out of the incoming bucket (different date) and the fuzzy probe
+    // picks it up from the in-memory pool (#1287). The wide ±48h prefetch
+    // window (MERGE_FUZZY_DEDUP=true in beforeEach) keeps it visible.
     mockEventFindMany.mockResolvedValueOnce([
       existingFuzzyRow({ runNumber: 786, title: "Annual Bash Trail" }),
     ] as never); // fuzzy probe candidate, but runNumber will conflict
@@ -3190,7 +3196,10 @@ describe("fuzzy ±48h cross-source dedup (#990)", () => {
 
   it("does NOT merge when locationName diverges sharply (different venues)", async () => {
     mockRawEventFind.mockResolvedValueOnce(null);
-    mockEventFindMany.mockResolvedValueOnce([] as never); // same-day empty
+    // Cache prefetch returns the candidate; getSameDayEvents filters it
+    // out of the incoming bucket (different date) and the fuzzy probe
+    // picks it up from the in-memory pool (#1287). The wide ±48h prefetch
+    // window (MERGE_FUZZY_DEDUP=true in beforeEach) keeps it visible.
     mockEventFindMany.mockResolvedValueOnce([
       existingFuzzyRow({ locationName: "Central Park" }),
     ] as never); // fuzzy probe candidate, location will conflict
@@ -3210,7 +3219,10 @@ describe("fuzzy ±48h cross-source dedup (#990)", () => {
 
   it("does NOT merge when startTime daypart conflicts (morning vs evening)", async () => {
     mockRawEventFind.mockResolvedValueOnce(null);
-    mockEventFindMany.mockResolvedValueOnce([] as never); // same-day empty
+    // Cache prefetch returns the candidate; getSameDayEvents filters it
+    // out of the incoming bucket (different date) and the fuzzy probe
+    // picks it up from the in-memory pool (#1287). The wide ±48h prefetch
+    // window (MERGE_FUZZY_DEDUP=true in beforeEach) keeps it visible.
     mockEventFindMany.mockResolvedValueOnce([
       existingFuzzyRow({ startTime: "10:00" }),
     ] as never);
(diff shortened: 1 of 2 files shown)


johnrclem:
@claude @codex review

johnrclem:
Pushed `16eccf95` addressing the review feedback. Three substantive bugs caught (CodeRabbit Major × 2, Gemini High, Codex P1, @claude Major × 3 — significant overlap):

**1. Cache write-back missing on UPDATE branch** (CodeRabbit / Gemini / @claude all caught this)

`getSameDayEvents` returns a freshly-filtered array, so `sameDayEvents[idx] = updated` only patched the local slice. The shared `ctx.sameDayCache` and `ctx.fuzzyPoolByKennel` still held the pre-update row reference. Concrete failure mode @claude described: source A and B both emit the same kennel+date in one batch, A high-trust-updates the canonical row's startTime/runNumber, B's lookup sees the STALE cached row, falls through all disambiguators, and creates a duplicate canonical record. Fix: patch both shared pools in place after the high-trust UPDATE, only when `!crossWindowMatch` (the cross-window path already invalidates the kennel cache).

I went with @claude's write-back approach over @gemini-code-assist's "always invalidate" approach because invalidation triggers an extra DB round-trip per UPDATE — that defeats the optimization #1287 is meant to deliver. Write-back keeps the cache fresh at zero query cost.

**2. `Promise.all` flush failing the whole batch** (CodeRabbit / Codex / @claude)

Pre-#1287 the equivalent `$executeRaw UPDATE "Kennel"` was inside the per-event try/catch — one transient DB error → `eventErrors++`, batch completed. The batched flush moved outside that boundary, so one rejected `kennel.updateMany` would reject the whole `processRawEvents` after all event mutations had already committed. Switched to `Promise.allSettled` with per-kennel `console.error` logging. `lastEventDate` is a non-critical UI cache field (used by the kennel directory's "recently active" filter); a missed update self-heals on the next scrape.

**3. Fuzzy rejection tests stopped exercising the rejection branches** (@claude)

Four negative tests (same-source guard, run-number mismatch, location divergence, daypart conflict) used the pre-#1287 two-mock pattern: empty same-day + populated fuzzy. The cache prefetch now consumes the FIRST mock (empty), so the fuzzy probe sees an empty in-memory pool and never reaches the rejection logic. Tests passed trivially even with the rejection branches deleted. Fix: consolidate to one mock seeding the candidate; the wide `MERGE_FUZZY_DEDUP=true` window in beforeEach keeps it visible to the fuzzy probe.

**New regression tests (2)**:
- "does not reject the whole batch when one kennel.updateMany flush fails" — locks in the `allSettled` semantics with a `console.error` spy.
- "writes a high-trust UPDATE back into the per-batch cache so later events see post-update state" — exercises the exact cross-source same-day match path that the original bug would have silently dropped.

300/300 merge tests + 6,313 full-suite tests pass; tsc + lint clean.

Skipping @claude's nitpick #4 (add `select` to prefetch queries) — low priority, already captured in the next-bottleneck list. Worth a follow-up but not for this PR.

PR #1361, opened 2026-05-10
refactor(merge): group eventBatch fields + narrow Event prefetch select

johnrclem:
## Summary

Two follow-ups deferred from PR #1359 review feedback, bundled in one PR:

### 1. `MergeContext` field grouping

`MergeContext` had 13 flat fields after #1287 added 4 (`batchDates`, `sameDayCache`, `fuzzyPoolByKennel`, `kennelMaxDates`). Collapsed those 4 into a sub-object:

```ts
ctx.eventBatch: {
  dates: Date[];
  sameDayByKennel: Map<string, EventRow[]>;
  fuzzyPoolByKennel: Map<string, EventRow[]>;
  kennelMaxDates: Map<string, Date>;
}
```

The four are tightly coupled (per-kennel cache + write buffers driving the same prefetch/flush pair) and only used by event-lookup paths. Mechanical rename across ~37 access sites; no behavior change.

### 2. Explicit `select` projection on the per-kennel event prefetch

```ts
const EVENT_CACHE_SELECT = { /* 30 fields */ } as const satisfies Prisma.EventSelect;
type EventRow = Prisma.EventGetPayload<{ select: typeof EVENT_CACHE_SELECT }>;
```

The cache prefetch in `ensureKennelEventCache` and the cross-window old-bucket refetch now use this select. Skips 6 scalar fields the merge pipeline never reads:

- `adminAuditLog` — Json column, grows unbounded on rows with cancel/uncancel history. The dominant savings on backfill scrapes that touch thousands of rows.
- `adminCancelledBy`, `adminCancellationReason`, `isManualEntry`, `submittedByUserId`, `updatedAt` — never accessed off cached rows.

Pattern matches the existing `RAW_EVENT_DEDUP_SELECT` for the RawEvent dedup prefetch (line ~385 in the same file). The single-use `FuzzyCandidate` type alias collapses to `EventRow` since the indirection no longer earns its keep.

## Why bundle these

Both were minor cleanups individually; together they're still small (+85 / −49, single file) and form a coherent "tidy-up the #1287 surface" PR.

## Test plan

- [x] `npx tsc --noEmit` clean (`EventRow` now narrows correctly across all access sites; `Event` from `createEventWithKennel` is structurally assignable)
- [x] `npm test src/pipeline/merge.test.ts` — 300/300 passing
- [x] `npm test` (full suite) — 6,313 passing
- [x] `/simplify` review applied 3 cleanups: trim verbose JSDoc on the new `EVENT_CACHE_SELECT` const (9 lines → 4) + the new `EventBatchState` interface (one-line per field), inline single-use `FuzzyCandidate` alias.

## Behavior change

None beyond the wire-payload narrowing (smaller `event.findMany` rows on cache prefetch + cross-window refetch). For an SDH3-style backfill (~70K rows fetched cumulatively), the realistic bandwidth savings is low single-digit MB on rows with empty `adminAuditLog`, but can dominate on rows with long cancel/uncancel histories.

### DIFF
--- a/src/pipeline/merge.ts
+++ b/src/pipeline/merge.ts
@@ -288,7 +288,43 @@ function shouldSkipReverseGeocode(sourceType: SourceType | null): boolean {
 }
 
 /** Per-batch state threaded through all merge helper functions. */
-type EventRow = Awaited<ReturnType<typeof prisma.event.findMany>>[number];
+/** Narrow select for the per-batch event cache. Skips columns the merge
+ *  pipeline never reads — notably `adminAuditLog`, a Json column that grows
+ *  unbounded on rows with cancel/uncancel history and would otherwise
+ *  balloon prefetch payloads on backfill scrapes. */
+const EVENT_CACHE_SELECT = {
+  id: true,
+  kennelId: true,
+  date: true,
+  dateUtc: true,
+  timezone: true,
+  runNumber: true,
+  title: true,
+  description: true,
+  haresText: true,
+  locationName: true,
+  locationStreet: true,
+  locationCity: true,
+  locationAddress: true,
+  latitude: true,
+  longitude: true,
+  startTime: true,
+  endTime: true,
+  cost: true,
+  trailLengthText: true,
+  trailLengthMinMiles: true,
+  trailLengthMaxMiles: true,
+  difficulty: true,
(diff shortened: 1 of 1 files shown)


johnrclem:
@claude @codex review

PR #1387, opened 2026-05-12
perf(merge): batch per-event RawEvent.update into one raw SQL

johnrclem:
## Summary

Sentry [JAVASCRIPT-NEXTJS-3](https://hashtracks.sentry.io/issues/JAVASCRIPT-NEXTJS-3) regressed (status: \`regressed\`, last seen today) after #1287 collapsed the read-side N+1s. The remaining offender is `prisma.rawEvent.update({ where: { id }, data: { processed: true, eventId } })` firing **per new event** in both branches of \`upsertCanonicalEvent\` (UPDATE-existing path at line ~1467, CREATE-new path at line ~1519).

The [Sentry sample event](https://hashtracks.sentry.io/issues/7375103156/events/3e80f0341f0a4b01891e385c8c3fc635/) shows 14 identical 3-5ms postgresql spans on a single SWH3 GCal scrape (87 total spans, 23 db spans, 14 of them this exact query).

## Approach

Mirrors #1287's `kennelMaxDates` batching:

1. **Defer** the writes into `ctx.eventBatch.processedRawEvents: Map<rawEventId, eventId>`. Filled at the upsert site (both CREATE and UPDATE branches).
2. **Flush** as ONE `$executeRaw` after the loop using a `FROM (VALUES …)` correlated update — Prisma's `updateMany` can't carry per-row payloads, only one shared `data`:
   ```sql
   UPDATE "RawEvent"
   SET "processed" = true, "eventId" = update_map."eventId"
   FROM (VALUES (id1, eid1), (id2, eid2), …) AS update_map(id, "eventId")
   WHERE "RawEvent"."id" = update_map.id
   ```
3. **Failure is logged not thrown** — the row recovers on the next scrape via the P2002 adopt-orphan path from #1286. Same self-healing semantic as the `kennelMaxDates` flush a few lines below.

## Sentry-visible win

N → 1 \`prisma:client:db_query\` spans per scrape for this query shape:
- SWH3-style GCal scrape (~14 new events/run): 14 → 1.
- SDH3-style backfill (~7K events): 7000 → 1.

Well inside the Postgres extended-query 65535-param limit (2 params/row → ≈32K-row implicit cap on this single flush).

## Tests

- **2 new regression tests** in the per-kennel batching describe block:
  - "batches every per-event RawEvent.update into one raw SQL after the loop" — asserts exactly one `$executeRaw` call with all (rawEventId, eventId) pairs in the SQL fragment.
  - "does not call $executeRaw when no new events are processed" — short-circuits on duplicate-only batches.
- **3 existing tests updated** — the OLD per-event wire-format assertions (`mockRawEventUpdate.toHaveBeenCalledWith(...)`) replaced with behavioral checks (`result.created`/`result.updated`). The new dedicated test covers the wire format end-to-end.

## /simplify review applied

- Trimmed verbose JSDoc on `processedRawEvents` (5 lines → 2) and the flush comment (11 lines → 6).
- Replaced brittle `mock.calls[0][1] as { values?: unknown[] }` assertions with `JSON.stringify(execCall).toContain(...)` — avoids coupling to `Prisma.sql`'s internal `{ strings, values }` shape (which isn't a public API).
- Surfaced `sourceId` in the failure log for Vercel log grep-ability.

## Test plan

- [x] `npm test src/pipeline/merge.test.ts` — 319/319 passing
- [x] `npm test` — 6,413 passing
- [ ] Production verify: after deploy, confirm the next SWH3 (or any new-event-heavy) scrape's Sentry trace shows 1 `$executeRaw` span instead of N `rawEvent.update` spans. Sentry's issue occurrence counter should plateau.

## Next N+1 candidates (not in scope)

Two reviewers flagged the same follow-up surface — both `eventLink.upsert` (per externalLink in `createEventLinks`, line 256) and `eventKennel.upsert` (per co-host secondary, line 1549) still fire per event. Less common today (<5% of scrapes have externalLinks or multi-kennel co-hosts), but same shape. Worth a follow-up issue.

Refs Sentry JAVASCRIPT-NEXTJS-3

### DIFF
--- a/src/pipeline/merge.test.ts
+++ b/src/pipeline/merge.test.ts
@@ -249,11 +249,9 @@ describe("processRawEvents", () => {
     const result = await processRawEvents("src_1", [buildRawEvent()]);
     expect(result.created).toBe(1);
     expect(mockEventCreate).toHaveBeenCalledTimes(1);
-    expect(mockRawEventUpdate).toHaveBeenCalledWith(
-      expect.objectContaining({
-        data: expect.objectContaining({ processed: true, eventId: "evt_1" }),
-      }),
-    );
+    // The mark-processed write is now batched into one raw SQL after the
+    // loop (see the "batches the mark-processed flush" describe block at
+    // the bottom of this file).
   });
 
   it("dual-writes primary EventKennel alongside the new Event (#1023 step 2)", async () => {
@@ -436,12 +434,9 @@ describe("processRawEvents", () => {
     mockEventFindMany.mockResolvedValueOnce([{ id: "evt_existing", trustLevel: 3 }] as never);
     mockEventUpdate.mockResolvedValueOnce({} as never);
 
-    await processRawEvents("src_1", [buildRawEvent()]);
-    expect(mockRawEventUpdate).toHaveBeenCalledWith(
-      expect.objectContaining({
-        data: expect.objectContaining({ processed: true, eventId: "evt_existing" }),
-      }),
-    );
+    const result = await processRawEvents("src_1", [buildRawEvent()]);
+    expect(result.updated).toBe(1);
+    // The mark-processed write is now batched (see the dedicated test below).
   });
 
   it("never writes locationCity for HARRIER_CENTRAL sources on create (#471)", async () => {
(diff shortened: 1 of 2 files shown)


johnrclem:
Pushed \`098be4e2\` addressing @codex's adversarial review.

**Fix applied (option 1 — rethrow):**
- Dropped the try/catch around the post-loop \`\$executeRaw\` flush.
- A flush failure now propagates as a rejected \`processRawEvents\` promise.
- New regression test "rethrows when the mark-processed flush fails so QStash retries on schedule" pins the behavior.

**One nuance with @codex's framing worth recording for the next reader:**

The review described the failure state as *"RawEvents stay in the database with \`eventId\` set but \`processed=false\`"*. Tracing the code, that specific state isn't actually producible — the buffered SQL writes both columns atomically (\`SET processed = true, eventId = update_map.eventId\`), so a flush failure leaves both columns unchanged. Fresh CREATE-branch rows started with \`processed=false, eventId=null\` (the Prisma defaults from \`rawEvent.create\`), so on flush failure they remain in that state — which is the orphan state, and the #1286 adopt-orphan path does heal it.

**But the underlying recommendation is still right.** The argument for rethrowing isn't "the swallowed state would be unrecoverable" — it's:

1. **Time-to-recovery on swallow:** next scheduled scrape (hours away).
2. **Time-to-recovery on throw:** QStash auto-retry (minutes).
3. Swallow also hides the failure from Sentry.

The new comment in \`merge.ts\` documents this more precisely so the next person doesn't repeat my mental-model error.

@codex if you can re-review and confirm the throw-path correctness, that'd close this out.

PR #1518, opened 2026-05-21
fix(schedule-rules): Pass 3 absorbs overlapping Pass 1 (#1492)

### LINKED ISSUE #1492: ScheduleRule pipeline dedup: Pass 1 ↔ Pass 3 collision when both emit HIGH-confidence rules for the same (kennelId, rrule)
## Background

Follow-up from [PR #1491](https://github.com/johnrclem/hashtracks-web/pull/1491) (which fixed the Pass 1 ↔ Pass 2 collision per #1486). After that fix lands, the only remaining duplicate-rule-pair surfaced by the audit query is `hebe-h3`:

| confidence | source | rrule | startTime | label |
|---|---|---|---|---|
| HIGH | STATIC_SCHEDULE | FREQ=MONTHLY;BYDAY=1SA | 15:00 | (null) |
| HIGH | SEED_DATA | FREQ=MONTHLY;BYDAY=1SA | 15:00 | Monthly |

Both rules render to the same display string (`"1st Saturday at 3:00 PM"` once the label resolves), so they previously joined as duplicates in the UI. The display-string dedup in PR #1491 (Fix 2) masks the symptom, but the redundant DB row remains.

## Why this is harder than Pass 1 ↔ Pass 2

Pass 2 (kennel display strings) emits MEDIUM/LOW confidence rules — when Pass 1 already has HIGH, Pass 2 is strictly lower-trust and can be dropped without losing anything.

Pass 3 (seed `scheduleRules`) emits HIGH-confidence rules that carry richer metadata: `label`, `validFrom`, `validUntil`, `displayOrder`. Hebe's Pass 3 row has `label="Monthly"`; Pass 1's STATIC_SCHEDULE row doesn't. Naïvely skipping Pass 3 when Pass 1 covers would lose that metadata.

## Proposed resolutions (design decision needed)

1. **Pass 3 wins, deactivate Pass 1's overlapping row.** When Pass 3 plans a rule with `(kennelId, rrule)` that matches an existing Pass 1 STATIC_SCHEDULE row, mark the Pass 1 row inactive (or delete it). Justification: seed `scheduleRules` is the deliberate authored shape; the STATIC_SCHEDULE source is the upstream inferred shape.
2. **Pass 1 wins, but merge metadata.** Keep the Pass 1 row, copy Pass 3's `label`/`validFrom`/`validUntil` into it, skip Pass 3 emit. Preserves the STATIC_SCHEDULE provenance.
3. **Both rows survive, but mark one as the display canonical.** Add a column like `isDisplayCanonical` and only the canonical row renders. More schema churn but preserves full audit trail.

(1) is probably right — Pass 3 represents explicit author intent — but worth thinking through whether losing STATIC_SCHEDULE provenance breaks any downstream consumer (event generation? travel projections?).

## Audit query (after #1491 lands)

```sql
SELECT k.\"kennelCode\", k.\"shortName\", COUNT(*) AS dup_count,
       STRING_AGG(sr.source::text || '/' || sr.confidence::text, ', ') AS sources
FROM \"ScheduleRule\" sr
JOIN \"Kennel\" k ON sr.\"kennelId\" = k.id
WHERE sr.\"isActive\" = true
GROUP BY k.id, k.\"kennelCode\", k.\"shortName\", sr.rrule, sr.\"startTime\"
HAVING COUNT(*) > 1
ORDER BY dup_count DESC, k.\"kennelCode\";
```

Currently returns 1 row (Hebe). Both rows are HIGH so the display dedup hides the symptom — this is mostly a DB-hygiene cleanup.

## Files likely affected

- \`scripts/backfill-schedule-rules.ts\` — \`runKennelSeedPass\` / \`planSeedRule\`
- Possibly \`scripts/backfill-schedule-rules.ts\` — deactivation logic at end of script

## Labels

\`bug\`, \`pipeline\`, \`area:schedule-rules\`, \`good first issue\` (small, well-scoped follow-up)

johnrclem:
## Summary

After PR #1491 fixed Pass 1↔Pass 2, the audit query still returned one duplicate row pair: `hebe-h3` had **both** a Pass 1 STATIC_SCHEDULE HIGH rule **and** a Pass 3 SEED_DATA HIGH rule for `(FREQ=MONTHLY;BYDAY=1SA, 15:00)`. Pass 3 carries `label="Monthly"` that Pass 1 cannot know; PR #1491's display-string dedup masked the symptom but left the redundant DB row in place.

**Resolution: Option A from the issue — Pass 3 wins, Pass 1 deactivates.**

- New helper `absorbOverlappingPass1Rule` runs inside `planSeedRule`. When a Pass 1 STATIC_SCHEDULE row with the same `(kennelId, rrule)` exists in `planned`:
  - Pass 3 inherits `startTime` / `anchorDate` from Pass 1 if the seed left them null (mirrors PR #1491's Pass 1↔Pass 2 enrichment direction).
  - Pass 3 adopts Pass 1's `lastValidatedAt` so the surviving row keeps tracking the scrape moment.
  - Pass 3 is marked `absorbsStaticSchedule: true`.
  - Pass 1 is spliced out of `planned`.
- `deactivateStaleRules` then finds the orphaned Pass 1 DB row active but missing from the plan and sets `isActive: false` (history preserved — script-time deactivation, not delete, per the issue's guidance).
- `applyUpserts` honors `absorbsStaticSchedule` by keeping `lastValidatedAt` in the UPDATE clause for the surviving SEED_DATA row.

## Downstream consumer audit

`prisma.scheduleRule.findMany` consumers in [src/lib/travel/search.ts:511](src/lib/travel/search.ts#L511) filter only on `isActive: true` and do not filter by `source`. The deactivated Pass 1 row drops out of all queries; the Pass 3 SEED_DATA row with the same rrule + adopted `lastValidatedAt` covers `scoreProjections` (which keys per-rule `lastValidatedAt`) identically. No other consumer reads ScheduleRule by provenance.

## Why `lastValidatedAt` for absorbed rows matters

Without the absorb flag, the Pass 3 row would inherit the SEED_DATA semantics (lastValidatedAt frozen at first-create) and the kennel would silently lose its "actively validated" signal in travel projections — `scoreProjections` uses `lastValidatedAt` per scheduleRuleId to decay confidence. The flag preserves the Pass 1 behavior exactly for the row that replaced it.

## Test plan

- [x] `npm test` — 6808 pass, 0 fail (57 backfill-schedule-rules tests, up from 56)
- [x] `npm run lint` — 0 errors
- [x] Synthetic Hebe reproduction asserts exactly one surviving row with `source: SEED_DATA`, `label: "Monthly"`, `absorbsStaticSchedule: true`, Pass 1's `lastValidatedAt`
- [x] Enrichment test: Pass 3 seed with null `startTime` + `anchorDate` inherits both
- [x] Over-fire guard: unrelated Pass 1 row (different kennel) untouched
- [x] applyUpserts spy: `absorbsStaticSchedule` SEED_DATA rules include `lastValidatedAt` in UPDATE

## Cleanup expectation post-merge

Re-run `tsx scripts/backfill-schedule-rules.ts --apply` (or wait for the next seed pass). The single hebe-h3 dup pair will collapse: the STATIC_SCHEDULE row flips `isActive: false`, the SEED_DATA row continues forward with `label="Monthly"` and refreshed `lastValidatedAt`.

Closes #1492

### DIFF
--- a/scripts/backfill-schedule-rules.test.ts
+++ b/scripts/backfill-schedule-rules.test.ts
@@ -683,6 +683,149 @@ describe("runKennelSeedPass", () => {
     expect(result).toEqual({ count: 0, skippedKennels: 0, skippedRules: 0 });
     expect(planned).toHaveLength(1); // unchanged
   });
+
+  // #1492: when Pass 1 already emitted a HIGH STATIC_SCHEDULE row for the same
+  // (kennelId, rrule), Pass 3 must absorb it — keep the SEED_DATA row (richer
+  // metadata), drop the STATIC_SCHEDULE row so `deactivateStaleRules` flips
+  // its DB row inactive. Hebe H3 is the production reproduction: both passes
+  // emit `(hebe-h3, FREQ=MONTHLY;BYDAY=1SA, 15:00)`; only the Pass 3 row
+  // carries label="Monthly".
+  it("(#1492) absorbs an overlapping Pass 1 STATIC_SCHEDULE row, surviving with richer Pass 3 metadata", async () => {
+    const pass1ValidatedAt = new Date("2026-05-01T00:00:00Z");
+    const planned: Parameters<typeof runKennelSeedPass>[1] = [
+      {
+        kennelId: "k_hebe",
+        kennelDisplay: "Hebe H3",
+        rrule: "FREQ=MONTHLY;BYDAY=1SA",
+        anchorDate: null,
+        startTime: "15:00",
+        confidence: "HIGH",
+        source: "STATIC_SCHEDULE",
+        sourceReference: "https://www.facebook.com/hebehash",
+        lastValidatedAt: pass1ValidatedAt,
+        notes: null,
+        label: null, // Pass 1 doesn't know the label
+        validFrom: null,
+        validUntil: null,
+        displayOrder: 0,
+      },
+    ];
+    const prisma = fakePrismaForSeed([
+      { id: "k_hebe", kennelCode: "hebe-h3", shortName: "Hebe H3" },
+    ]);
+    const seeds = [
+      {
+        kennelCode: "hebe-h3",
+        scheduleRules: [
+          { rrule: "FREQ=MONTHLY;BYDAY=1SA", startTime: "15:00", label: "Monthly" },
+        ] satisfies KennelScheduleRuleSeed[],
+      },
+    ];
+    await runKennelSeedPass(prisma, planned, {}, seeds);
+
+    // Exactly one survivor: the Pass 3 SEED_DATA row with absorbsStaticSchedule set.
+    expect(planned).toHaveLength(1);
+    const [survivor] = planned;
+    expect(survivor).toMatchObject({
+      kennelId: "k_hebe",
+      rrule: "FREQ=MONTHLY;BYDAY=1SA",
+      startTime: "15:00",
+      confidence: "HIGH",
+      source: "SEED_DATA",
+      label: "Monthly", // Pass 3 metadata preserved
+      absorbsStaticSchedule: true,
+    });
+    // Pass 1's scrape moment carries over so the surviving row keeps tracking
+    // validation freshness (see applyUpserts + travel scoreProjections).
+    expect(survivor.lastValidatedAt).toBe(pass1ValidatedAt);
+  });
+
+  // #1492 enrichment: Pass 3 inherits null fields from the absorbed Pass 1 row.
+  // A seed that omits startTime should still surface the Pass 1 source config's
+  // startTime, mirroring the Pass 1 ↔ Pass 2 enrichment behavior added in #1491.
+  it("(#1492) inherits Pass 1's startTime + anchorDate when the Pass 3 seed left them null", async () => {
+    const planned: Parameters<typeof runKennelSeedPass>[1] = [
+      {
+        kennelId: "k_x",
+        kennelDisplay: "X",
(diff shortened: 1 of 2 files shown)


johnrclem:
**Codex adversarial review applied** — two HIGH findings on the absorption path, both addressed in 5992ba9.

- **HIGH #1**: `lastValidatedAt: null` on Pass 1 (scrape never succeeded) was silently combined with Pass 3's synthetic `new Date()` default + `absorbsStaticSchedule: true` → row appeared "freshly validated" on every backfill run with no actual scrape behind it.
- **HIGH #2**: shape disagreement (Pass 1 `startTime=18:00` vs seed `startTime=19:00`) had Pass 3 win on identity (correct) but still inherit Pass 1's `lastValidatedAt` — a validation timestamp earned by a different shape than what gets stored.

Fix adds two explicit gates inside `absorbOverlappingPass1Rule`:
1. Detect `startTimeConflict` / `anchorDateConflict` BEFORE the null-inheritance step.
2. Set `absorbsStaticSchedule = true` only when **shapes agree AND Pass 1 has a non-null timestamp**. Otherwise still splice (Pass 3 wins identity, per the issue's Option A) but keep standard SEED_DATA semantics — travel projections then decay confidence naturally until a future scrape confirms.

Three new tests cover the new gate paths; existing tests still pass (they exercise the agree-and-validated case).

(Original Codex output: needs-attention → both HIGH defects above. No MEDIUM findings.)

johnrclem:
@claude @codex review

PR #1811, opened 2026-05-29
docs(kennel-onboarding): fold ONH3 retro fixes into prompt + Bali patch

johnrclem:
See commit message. Source-of-truth retro is committed under `handoff-retros/` (outside `handoffs/` so the copy-newest helper skips it). Bali Hash 2 implementation can paste the updated handoff (re-run `scripts/copy-newest-handoff.sh` first) and won't hit the inferCountry gap that bit ONH3.


johnrclem:
@claude @codex review

PR #1842, opened 2026-05-30
fix(hareline): cap upcoming queries to stay under unstable_cache 2MB limit

johnrclem:
## Summary

- **Root cause:** the global unfiltered upcoming payload (~5 272 events × ~906 bytes ≈ 4.8 MB) silently exceeded Next.js's 2 MB `unstable_cache` per-entry ceiling, so every anonymous hareline request re-ran a live Prisma query instead of hitting the data cache.
- Added `UPCOMING_GLOBAL_LIMIT = 1500` (~1.3 MB, ~30% headroom) and `UPCOMING_KENNEL_LIMIT = 2000` (~1.8 MB) — kennel-filtered paths get a higher budget since subscribers expect their kennel's full upcoming schedule.
- Replaced the old `isPast`-only `take` guard with a clean three-way ternary covering all paths.
- Encoded both limits in the `unstable_cache` static key so a future constant change auto-busts stale entries instead of silently serving old-sized payloads for up to an hour.
- Fixed sort-before-slice in `kennelIds` normalisation: the old code sliced before sorting, making which IDs were dropped for >50-kennel inputs depend on call-site argument order.
- Added `eventLabel` to `HarelineListEvent` — the field was selected from the DB and mapped into the return object but absent from the interface, so `EventCard` always read `undefined` despite the value being present.

## Trade-off acknowledged

The global cap means kennels/regions whose next event falls beyond position 1 500 (date-ascending) are absent from the FilterBar kennel dropdown and RegionQuickChips counts, which derive from the returned payload. This is documented in the `UPCOMING_GLOBAL_LIMIT` comment. The kennel directory remains the authoritative discovery surface.

## Test plan

- [x] `npm test -- src/app/hareline` — 8/8 pass
- [x] `npx tsc --noEmit` — no new errors (pre-existing implicit-`any` from missing generated Prisma types unchanged)
- [ ] After deploy: confirm dev-server log no longer shows `"Failed to set Next.js data cache … items over 2MB"` for the global upcoming path


johnrclem:
@claude review

Issue #1866, opened 2026-05-30
Oregon H3 — Profile bundle: foundedYear, scheduleFrequency, scheduleNotes, description


johnrclem:
## Summary

Oregon H3 kennel profile is missing several fields that the source website publishes openly. The fields below have a visible home on the kennel page today.

## Fields to populate

| Field | Current | Proposed | Source |
| --- | --- | --- | --- |
| `foundedYear` | null | `1987` | https://www.oregonhhh.org/welcome.html — "Welcome to the Oregon Hash" was first chanted on May 17, 1987 |
| `scheduleFrequency` | "Biweekly" | "Biweekly + Full Moon" | https://www.oregonhhh.org/welcome.html — "Runs held on the 2nd & 4th Saturdays and the Full Moon" |
| `scheduleNotes` | "Bi-weekly Saturdays plus full moon runs" | "2nd & 4th Saturdays at 1pm plus Full Moon evening runs. Hotline 866-656-5477 for run info." | https://www.oregonhhh.org/welcome.html |
| `description` | "Oregon's flagship kennel. Bi-weekly Saturday afternoon trails plus full moon evening runs in the Portland metro area." | (proposed) "Founded 17 May 1987 by Wrong Way Corrigan — THE Mother Hash of Oregon. Biweekly Saturday afternoon trails plus Full Moon evening runs in the Portland metro area. Tyrant I and founder: Mark 'Wrong Way Corrigan' Cook." | https://www.oregonhhh.org/welcome.html + https://www.oregonhhh.org/tyrants.html |

## Notes (no schema column today; flagged elsewhere or skipped)

- `logoUrl` — source www.oregonhhh.org has no `og:image` and no obvious header logo; HashRego `/kennels/OH3/` is actually Okinawa H3 (shortcode collision — `OH3` on HashRego = Okinawa, `OHHH` = Ottawa). Leaving `logoUrl` null.
- `facebookUrl`, `instagramHandle`, `xHandle` — source page only lists a Yahoo Groups URL (`sports.groups.yahoo.com/group/OH3/`), which is dead (Yahoo Groups shut down in 2020). No active social properties advertised on the website. Skipping.
- `contactEmail` — none published; the "Hash/Trash Talk" link points to the (dead) Yahoo group. Hotline (866-656-5477) is a phone, not an email, and we have no `Kennel.phone` column.
[...]

Issue #1818, opened 2026-05-29
OCHHH — Profile bundle: scheduleFrequency (Monthly→Biweekly), plus notes on website (dead) and foundedYear (~1979 soft signal)


johnrclem:
## Missing / incorrect profile fields

Sources: Google Calendar `hash.org_gr8mpprvpgpiihhkfj0dd0ic4k@group.calendar.google.com` (`X-WR-CALNAME: OCHHH`, `X-WR-CALDESC: Orange County Hash House Harriers`). The kennel’s historical website `ochhh.org` no longer resolves (NXDOMAIN) and `ochhh.com` is a parked / lander domain — nothing useful to set `website` to.

| Field | Current HT | Proposed value | Source signal |
|---|---|---|---|
| `scheduleFrequency` | `Monthly` | `Biweekly` (or roughly “every 2–3 weeks on Saturday”) | 2026 cadence in the source calendar: 1/10, 2/14, 3/7, 3/14, 4/4, 4/18, 5/2, 6/6, 7/11, 8/1, 8/22, 9/5, 9/19, 10/3, 10/31, 11/14 — about 2 runs/month, never monthly. Long-run average from the ICS: ~30 events/yr in 2012–2021 (closer to weekly with skips), ~17–24/yr in 2022–2026 (clearly biweekly-ish). |
| `scheduleDayOfWeek` | `Saturday` | `Saturday` (no change — confirmed) | Every 2026 event is on Saturday. |
| `scheduleTime` | `10:00 AM` | `10:00 AM` (no change — confirmed by per-event start times) | All current events have `startTime=10:00`. |

## Other gaps (not enough signal to set yet)

- `website`: leaving null. Historical site `ochhh.org` returns `ERR_NAME_NOT_RESOLVED`; `ochhh.com` is a parking page. No replacement site found.
- `foundedYear`: source-text hint is `“Since 1979 Every 2nd Saturday of July”` in the OCHHH-Train-Moon (AMTRAK Mooning Run) event description — a long-running OCHHH tradition. This is suggestive of a 1979-era kennel founding but the text technically describes the *tradition*, not the *kennel*. Not setting `foundedYear` per the math-only rule, but worth a note for whoever has direct confirmation.
- `hashCash`: inconsistent across event descriptions ($5 typical, $2 BYOB on Pookie’s trail, $0 on Rumpy’s Kinda-Virtual). Leaving null is correct; could note “typically $5” once a `hashCash` UI accepts that nuance.
[...]

PR #1887, opened 2026-06-01
chore(scripts): record O-cluster A profile-override one-shot (#1818, #1866)

johnrclem:
## Summary

Records the one-shot profile-override wrapper used to apply the OCHHH (#1818) and Oregon H3 (#1866) profile changes from [#1875](https://github.com/johnrclem/hashtracks-web/pull/1875) to prod.

Those fields (`scheduleFrequency`, `scheduleNotes`, `description`) were already **non-null** in prod, and `prisma/seed.ts` does a **fill-only** merge for existing kennels (seed.ts:296-301) — it never overwrites a populated column. So the seed couldn't apply them; the override runner is the only path. This wrapper mirrors the existing `cleanup-new-kennel-profiles.ts` / `cleanup-cycle6-profile-overrides.ts` pattern (drift-guarded, idempotent, dry-run by default).

**Already executed against prod** (post-merge): both kennels updated, verified. This commit is the audit-trail record so the change is reproducible.

No runtime/app code; one new script only.

### DIFF
--- a/scripts/cleanup-ocluster-a-profiles.ts
+++ b/scripts/cleanup-ocluster-a-profiles.ts
@@ -1,0 +1,66 @@
+/**
+ * One-off profile overrides for the O-cluster A bundle (#1818 OCHHH, #1866 Oregon H3).
+ *
+ * These fields are already non-null in prod with the old value, so the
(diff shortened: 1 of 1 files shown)


johnrclem:
@claude review

PR #1919, opened 2026-06-02
fix(merge): lower-trust enrichment backfills coords/map URL, not just venue name

johnrclem:
## Summary
The merge pipeline's lower-trust enrichment branch (`src/pipeline/merge.ts`, the `else` of `if (ctx.trustLevel >= existingEvent.trustLevel)`) backfilled NULL canonical-Event fields from a lower-trust source — `description`, `eventLabel`, `haresText`, `locationName`, `locationStreet`, `startTime`, the trail-length bundle, `difficulty`, `trailType`, `dogFriendly`, `prelube`, `endDate` — but **not** `latitude`, `longitude`, or `locationAddress` (map URL).

That asymmetry meant a lower-trust source could fill a venue's **name** but its map **pin** was silently dropped: if a higher-trust source created an Event with no location and a lower-trust source carried the venue + coords, the event rendered without a map. A **Codex adversarial review** caught this during the NSWHHH onboarding ([#1917](https://github.com/johnrclem/hashtracks-web/pull/1917)), which worked around it by making the coord-bearing website source out-trust the sheet. This closes the general gap so dual-source kennels can rely on enrichment instead of trust-ordering gymnastics.

## Change
The enrichment branch now resolves coords and fills `latitude`/`longitude`/`locationAddress` (+ reverse-geocoded `locationCity`) when the canonical has them null:
- **Reuses** the same helpers the sibling full-update branch uses — `resolveCoords`, `shouldSkipReverseGeocode`, `reverseGeocode`, `suppressRedundantCity`, `sanitizeLocationUrl`.
- **Fill-only** — never overwrites a higher-trust source's coords (guarded on `latitude == null && longitude == null` / `!locationAddress`). The map-URL fill is independent of coords (a source may carry a maps link even when coords don't resolve).
- **Skips the async resolve** entirely unless the source carries a coord signal (direct lat/lng, a maps URL, or a venue to geocode); `existingCoords` is omitted from the call since the null-guard means there's nothing cached to preserve.
- `locationCity` mirrors the full-update reverse-geocode (skipped for canonical-location source types, only when null).

## Tests (`src/pipeline/merge.test.ts`)
- **Lower-trust source fills coords** on a coordless canonical Event → `latitude`/`longitude`/`locationAddress` land.
- **Lower-trust source does NOT overwrite** existing coords → only the genuinely-null field (`description`) is filled; lat/lng/`locationAddress` untouched.
- Updated the existing "NSWHHH dual-source" test's comment (the branch *does* backfill coords now).

## Verification
`npx tsc --noEmit` ✅ · `npm run lint` ✅ · `npm test` ✅ (8379 passed; 370 in merge.test.ts). `/simplify` ran (2 cleanups applied: omit the inert `existingCoords` object; add the coord-signal pre-guard).

Note: a deeper refactor — extracting a shared `deriveCoordCity` helper used by the full-update, enrichment, and create branches — was flagged by the review as a reasonable future cleanup but deliberately left out of this focused fix.

### DIFF
--- a/src/pipeline/merge.test.ts
+++ b/src/pipeline/merge.test.ts
@@ -510,12 +510,11 @@ describe("processRawEvents", () => {
 
   it("higher-trust enrichment source writes coords onto a no-coords canonical event (NSWHHH dual-source)", async () => {
     // NSWHHH ships two sources: the hareline sheet (trust 7, no venue/coords)
-    // and the website (trust 8, venue + map coords). If the sheet's raw merges
-    // first, the canonical Event is created without coordinates. The website
-    // must OUT-trust the sheet so it takes the full-update path and writes
-    // lat/lng — the lower-trust enrichment branch backfills locationName but
-    // NOT locationAddress/latitude/longitude (merge.ts ~L1668), so a lower-trust
-    // website would have its map pin silently dropped (Codex adversarial review).
+    // and the website (trust 8, venue + map coords). When the higher-trust
+    // website merges, it takes the full-update path and writes lat/lng. (As of
+    // the enrichment-coords fix the lower-trust path also backfills coords — see
+    // the "lower-trust enrichment backfills coords" test below — but the
+    // higher-trust write is still exercised here.)
     mockSourceFind.mockResolvedValueOnce(sourceRow({
       trustLevel: 8,
       type: "HTML_SCRAPER",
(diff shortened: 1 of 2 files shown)


johnrclem:
Addressed the review feedback in a4bb9a47:

**Codex adversarial review (high) — venue/pin mismatch.** Valid catch: the coord backfill only checked that the canonical's coords were null, not that the lower-trust source's *venue* matched the existing higher-trust venue name — so a fuzzy/stale match could pin a different venue's coords under the higher-trust name (silent map/name mismatch). Added a **venue-compatibility gate**: coords + map-URL backfill only when the canonical has no venue yet (name + pin then fill from the same source, agreeing by construction) **or** `sanitizeLocation(event.location) === existingEvent.locationName`. Strict equality is intentionally conservative — on any doubt, skip rather than risk a wrong pin. Added the regression test Codex suggested (existing higher-trust venue + null coords + a lower-trust match with a *different* venue → no backfill).

**CodeRabbit nitpick — test only exercised direct lat/lng.** Added a test that derives coords via `geocodeAddress` (venue string, not raw lat/lng) and asserts the map-URL backfill + the reverse-geocoded `locationCity` fill. (`extractCoordsFromMapsUrl` is stubbed to null in this test file, so the URL→coords path can't be exercised there; geocode is the established pattern — see the `dropCachedCoords` test.)

372 merge tests pass; full suite green (8381).

Issue #1976, opened 2026-06-04
Paris H3 — Duplicate event for R*n 1136 (admin seed + adapter), same pattern on Sans Clue R*n 1192


johnrclem:
### Paris H3 — Duplicate event for R*n 1136 (admin seed + adapter), and same pattern on Sans Clue R*n 1192

**Source**: https://www.meetup.com/parish3-schhh/events/
**Suspected adapter**: `MEETUP`

**Paris H3 R*n 1136 (2026-06-13)** is stored twice:

| Event ID | createdAt cuid prefix | Notes |
|---|---|---|
| `cmpxbb2c000025ihnuj4rqwhr` | `...0000` (admin seed) | Likely created at kennel-setup time; identical title/date/sourceUrl to row below |
| `cmpxbq0ux000404jvo0v6m5t9` | `...0004` (adapter pull) | Clean MEETUP adapter row |

Both have title `Paris H3 R*n 1136 | TBD`, date `2026-06-13T12:00`, startTime `14:00`, sourceUrl meetup.com/parish3-schhh.

**Cross-kennel parallel**: The sister kennel Sans Clue H3 has the *exact same* seed-then-scrape duplication on R*n 1192 (`cmpxbb2eg00035ihnhh6nxf8j` seed vs `cmpxbq0kk000204jvulyzksit` adapter). Both kennels were seeded with their next-upcoming run on the same admin pass (`cmpxbb` prefix), then the adapter pulled the same row again with a fresh cuid (`cmpxbq` prefix) because the upsert key (date + sourceUrl) didn't match.

**Proposed fix**: delete the older seed cuid in each kennel (`cmpxbb2c000025ihnuj4rqwhr` for Paris, `cmpxbb2eg00035ihnhh6nxf8j` for Sans Clue). Same root-cause family as #1962 (Asunción Run #120), #1967 (Auckland H3 2026-06-08).

johnrclem:
## Not reproducible — the "seed event IDs" are Kennel IDs, not Events

Verified against prod. The two IDs reported as a "seed vs adapter" duplicate pair are not two events — the first ID in each pair is that **kennel's own `Kennel.id`**:

| Reported "seed event ID" | What it actually is |
|---|---|
| `cmpxbb2c000025ihnuj4rqwhr` | `Kennel.id` of **paris-h3** (Paris H3) — not an Event |
[...]

Issue #1521, opened 2026-05-21
Capital H3 — Source rows missing a Location field cause the Hare cell to be stored as Location (3 events)


johnrclem:
The HTML_SCRAPER for sporty.co.nz/capitalh3 is mis-handling source rows that don't supply all four en-dash-separated fields. Result: the Hare cell ends up as the Location.

**Source notice format** (verbatim from https://www.sporty.co.nz/capitalh3):
```
2326 – 18 May 2026 – The Bond Sports Bar – Geestring
2327 – 25 May 2026 – The Bridge Bar – Scrac Thing
2328 – 1 Jun 2026 – 5pm? Kings B'day – Hare required! –
2329 – 8 Jun 2026 – Hare required! –
2330 – 15 Jun 2026 – Hare required! –
```

Rows 2326 and 2327 have 4 fields (run / date / location / hare) and extract cleanly. Rows 2329 and 2330 have only 3 fields (run / date / hare-placeholder) — the source omits the Location cell entirely. Row 2328 has 4 fields but the Location cell is the holiday note "5pm? Kings B'day" rather than a real venue.

**HashTracks output:**

| Run | HashTracks Location | Real source intent |
|---|---|---|
| 2328 (https://www.hashtracks.xyz/hareline/cmpagmdk80006wehnnt7gkdwm) | "5pm? Kings B'day" | Holiday note / start-time override, not a venue |
| 2329 (https://www.hashtracks.xyz/hareline/cmpagmduh0008wehngobjt8xu) | "Hare required! -" | Hare placeholder; no location supplied |
| 2330 (https://www.hashtracks.xyz/hareline/cmpagme3l000awehnk6h0dwae) | "Hare required! -" | Hare placeholder; no location supplied |

**Fix hypothesis:** the adapter is splitting on `–` (en-dash) and assuming position 2 = location, position 3 = hare regardless of field count. It needs to:
1. Detect known placeholder tokens ("Hare required!", "TBA") and not store them in any field — they're the source's way of saying "this slot is empty".
2. Recognize when a row has 3 fields vs 4: a 3-field row almost always means "no location yet" (because the hare-placeholder is the last cell), not "no hare".
3. For row 2328 specifically: "5pm? Kings B'day" doesn't match any address pattern (no digits-then-street-word). It's a candidate for the description/notes field, not Location — and the "5pm" prefix is a per-event start-time override that should land in startTime when #504 ships.

[...]

Issue #530, opened 2026-04-08
Merge: time-shaped fields can't be cleared by sources (startTime/endTime/cost)


johnrclem:
Surfaced by Codex adversarial review on #529.

## Problem

The merge update path in \`src/pipeline/merge.ts\` writes time-shaped fields with the \`?? existingEvent.X\` pattern:

\`\`\`ts
startTime: event.startTime ?? existingEvent.startTime,
endTime:   event.endTime   ?? existingEvent.endTime,
cost:      event.cost      ?? existingEvent.cost,
\`\`\`

Because adapters normalize \"missing\" and \"explicitly removed\" both to \`undefined\`, there is no way for a source to clear a value once it has been populated. If a source removes an end time, switches to an all-day event, drops a price, or fixes a typo'd start time, the stale canonical value sticks around indefinitely.

By contrast, fields like \`description\` and \`location\` use a \`!== undefined\` gate so adapters can send \`null\` to clear.

## Scope

This is the long-standing \`startTime\` convention; \`endTime\` and \`cost\` (added in #529) inherited it for consistency rather than diverging. Worth fixing all three together rather than just the new fields.

## Proposed approach

1. Allow \`RawEventData.startTime\` / \`endTime\` / \`cost\` to be \`string | null\` (where \`null\` = explicit clear, \`undefined\` = source didn't provide)
2. Switch the merge update path for those three fields to the same \`!== undefined\` gating used by description/location
3. Audit each adapter so that \"the value used to be there and is now gone\" emits \`null\`, not \`undefined\`
4. Add a regression test: a source provides start/end/cost, then a re-scrape omits them — canonical Event should be cleared

Splitting from #529 because it touches the existing \`startTime\` semantics and warrants its own review.

PR #1987, opened 2026-06-05
fix(merge): field-level enrich fill for cost + endTime (#1950)

### LINKED ISSUE #1950: Merge field-fill is whole-record trust-gated, not field-level: a higher-trust source lacking a field blocks a lower-trust source from filling it
## Summary
When multiple sources co-cover the same `(kennelId, date)`, the merge update path is gated by whole-record trust ranking, not per-field completeness. A lower-trust source's RawEvent **merges onto** the canonical (its `eventId` is set, so it appears in the linked-raws list) but its field updates — including the `cost` tri-state — are **skipped** when a higher-trust source owns the canonical, *even if that higher-trust winner doesn't provide the field*.

## Where
- Gate: `isHigherOrEqualTrust = ctx.trustLevel >= existingEvent.trustLevel` in `src/pipeline/merge.ts` (~line 701).
- Field write it blocks (example): `cost` update at `src/pipeline/merge.ts` (~line 1565): `...(event.cost !== undefined ? { cost: event.cost ?? null } : {})`.

## Concrete evidence (prod, found verifying #1126 / PR #1945)
EWH3 #1521 (2026-06-04) canonical `cmlrh0fr3003u04jxkeqk0ewb`, `cost = null`. Linked RawEvents:

```
Hash Rego (trust 8)               cost=$10   ← merged onto the event, but update skipped
EWH3 WordPress Trail News (trust 9) cost=—   ← owns the canonical; has no cost
EWH3 Google Calendar (trust 7)    cost=—     (×8)
```

Because the canonical is owned by the trust-9 WordPress source, Hash Rego's `8 >= 9` check is false, so its `$10` never writes → the event shows no cost. By contrast "EWH3 Pub Crawl" (no trust-9 competitor; Hash Rego is highest) correctly shows `$45`. So Hash Rego cost lands on EWH3 events that are GCal-only (trust 7 < 8) but **not** on ones the trust-9 WordPress source also covers.

## Impact
- Currently narrow: of 53 upcoming EWH3 events, exactly **1** (#1521) is affected (the only event that's both trust-9-co-covered *and* Hash-Rego-costed).
- But it's a **general mechanism**: it recurs for any per-event field (`cost`, `hares`, `location`, etc.) supplied only by a lower-trust enrichment source when a higher-trust source co-covers the date without that field. The "wire up cost from Hash Rego" work (#1126) is otherwise complete and verified — this is the remaining edge.

## Options (each has broad blast radius — decide deliberately)
1. **Field-level fill-if-empty** — let an *empty* canonical field be filled by any source regardless of trust, while keeping trust ranking for *conflicting* (both-present) values. Most correct for enrichment, but risks low-trust sources injecting data into empty fields.
2. **Re-rank source trust** — e.g. raise Hash Rego above WordPress. Risks the higher-trust source losing fields (title/hares) it legitimately *should* win. Not recommended.
3. **Accept the gap** — document that low-trust enrichment fields don't surface when a higher-trust source co-covers.

## Not in scope for adapter/alias/config PRs
This is a shared merge-pipeline semantics change, not an adapter/alias/source-config fix. Surfacing as its own issue so it can be scoped and reviewed independently.

Discovered during post-merge verification of #1931 → #1945 (Hash Rego EWH3 cost).

johnrclem:
## Summary
Fixes the field-level trust-gating gap in `src/pipeline/merge.ts` reported in #1950.

The lower-trust **enrichment branch** in `upsertCanonicalEvent` already fills NULL canonical fields from a lower-trust co-covering source (description, hares, location, coords, startTime, trail-length, difficulty, …). It had silently omitted exactly **two** fields — `cost` and `endTime` — the only enrichable fields the higher-trust full-update branch writes that weren't mirrored here. So a lower-trust source's cost/end-time was dropped whenever a higher-trust source owned the canonical without those fields.

### Live evidence (prod, verified)
Canonical `cmlrh0fr3003u04jxkeqk0ewb` (EWH3 #1521, 2026-06-04): `cost=NULL`, trustLevel 9 (WordPress owns it, no cost). Linked Hash Rego raw (trust 8) carries `cost=$10` — never wrote because `cost` wasn't in the enrich branch. After this fix the next EWH3 merge fills `$10`.

## Changes
- **Enrich-branch fills** for `cost` + `endTime` — fill-only (`!existingEvent.X && event.X`), mirroring the ~13 sibling fills. The tri-state explicit-clear (`null`) stays owned by the higher-trust branch; `endTime` has no `dateUtc` coupling so no recompose.
- **`patchSharedEventCaches` helper** extracted from the higher-trust update branch and now called from the enrich branch too. The enrich UPDATE previously patched only the local `sameDayEvents` copy, not the shared per-batch pools — so a second same-batch row re-matching the canonical could clobber an earlier fill. (Caught by `/codex:adversarial-review`.)

## Tests
- `it.each` fill / no-override for `cost` + `endTime` (enrich branch).
- Same-batch **no-clobber** regression — verified to fail without the shared-cache patch.
- Fingerprint-participation guards for `cost`/`endTime`.
- Explicit-clear tri-state already covered by the existing #530 suite (untouched higher-trust branch).
- `npm run lint && npx tsc --noEmit && npm test` → green (8482 passing).

Post-deploy live-verify: re-scrape EWH3, re-query `cmlrh0fr3003u04jxkeqk0ewb` → expect `cost=$10`.

> Note: the sibling duplicate-event issues (#1976/#1984/#1962/#1967) are being resolved separately as not-reproducible — every "seed event ID" in those reports is actually that kennel's own `Kennel.id`, and prod has zero duplicates for those kennels.

Closes #1950

### DIFF
--- a/src/pipeline/fingerprint.test.ts
+++ b/src/pipeline/fingerprint.test.ts
@@ -75,6 +75,30 @@ describe("generateFingerprint", () => {
     expect(generateFingerprint(baseline)).toBe(generateFingerprint(withExplicitUndefined));
   });
 
+  // #1950 stability guard — the lower-trust enrich branch now fills cost +
+  // endTime onto the canonical. Both fields already participate in the
+  // fingerprint (cost/endTime are WS6 tokens), so a RawEvent that newly carries
+  // them re-fingerprints and the canonical UPDATE persists the enriched value.
+  // These assert that wiring stays intact: distinct values fork the fingerprint,
+  // and a baseline event with neither field is unaffected (no re-merge churn).
+  it.each([
+    ["cost", { cost: "$10" } as const],
+    ["endTime", { endTime: "21:30" } as const],
+  ])("different %s values produce different fingerprints (#1950)", (_field, override) => {
+    const a = generateFingerprint(buildRawEvent());
+    const b = generateFingerprint(buildRawEvent(override));
+    expect(a).not.toBe(b);
+  });
+
+  it.each([
+    ["cost", { cost: undefined } as const],
+    ["endTime", { endTime: undefined } as const],
+  ])("%s=undefined leaves the baseline fingerprint unchanged (#1950)", (_field, override) => {
+    expect(generateFingerprint(buildRawEvent())).toBe(
+      generateFingerprint(buildRawEvent(override)),
+    );
+  });
+
   // #1624 — eventLabel must participate in the fingerprint so a label edit
   // (e.g. "Birthday" → "Pink Moon") re-fingerprints and the canonical UPDATE
   // fires; otherwise the dedup map silently swallows the new label.
--- a/src/pipeline/merge.test.ts
+++ b/src/pipeline/merge.test.ts
@@ -635,6 +635,135 @@ describe("processRawEvents", () => {
     expect(data).not.toHaveProperty("locationAddress");
   });
 
+  // #1950 — endTime + cost were the last two enrichable fields the lower-trust
+  // enrich branch omitted (the higher-trust full-update branch always wrote
+  // them). A lower-trust co-covering source must fill them when the higher-trust
+  // owner left them NULL — EWH3 #1521: Hash Rego (trust 8) carried cost "$10"
+  // but the trust-9 WordPress source owned the canonical with cost=null, so the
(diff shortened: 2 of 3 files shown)


johnrclem:
@claude @codex review

Issue #2066, opened 2026-06-09
Latent time-bomb tests: hardcoded fixture dates + date-window filter (32 adapter test files, detonating 2027–2028)


johnrclem:
## Summary

Adapter `fetch()` tests that combine **hardcoded absolute fixture dates** with the adapter's **relative date-window filter** (`buildDateWindow` / `applyDateWindow`, typically ±`days`) and a **non-empty / exact-count events assertion** are time-bombs: once the wall clock passes the fixture dates by more than the window, the events age out and the assertion goes red — redlighting CI for every PR until fixed.

This is the same defect that hit `atlanta-hash-board.test.ts` (fixed in `53d19f11`, "de-bomb atlanta-hash-board fetch test").

The **imminent** instances were fixed in the companion PR:
- `atlanta-hash-board.test.ts` (already landed)
- `dublin-hash.test.ts`, `sdh3.test.ts` (±90-day windows — detonating mid-June 2026)
- `sdh3.test.ts` (±365), `squarespace-events.test.ts` (late-2026 / 2027)

This issue tracks the **remaining latent instances**, which detonate across **2027–2028** (not imminent).

## How this list was generated

Ran the adapter suite with the system clock advanced to `2028-06-09` (Date-only fake, set at vitest `setupFiles` top level so module-load-time fixtures also see the future date — this removes false positives from tests that use *relative* fixtures). Tests still failing under that clock use hardcoded absolute dates that will age out.

```ts
// setupFile
import { vi } from "vitest";
vi.useFakeTimers({ toFake: ["Date"] });
vi.setSystemTime(new Date("2028-06-09T12:00:00Z"));
```

Caveats: the list is a strong starting catalogue, not exhaustive — fixtures dated beyond 2028 won't show up at this horizon, and a few entries may be relative fixtures built inside `beforeEach`. Re-run and inspect each before fixing.

## Recommended fix (consistent with the imminent PR)

Two sanctioned approaches (pick per file):
1. **Relative fixture dates** — compute fixture dates from `new Date()` so the in-window event never ages out (see the `relativeDate` / `isoDaysAgo` helpers in `dublin-hash.test.ts` / `atlanta-hash-board.test.ts`).
[...]

PR #2069, opened 2026-06-09
test(adapters): de-bomb imminent hardcoded-fixture-date tests

johnrclem:
## Summary

De-bombs the **imminent** sibling instances of the `atlanta-hash-board` time-bomb (fixed earlier in `53d19f11`): adapter `fetch()` tests that pin **absolute fixture dates** which age out of the adapter's relative `buildDateWindow`/`applyDateWindow` (±`days`) window, then assert events survive — redlighting CI for every PR once the wall clock passes them.

`atlanta-hash-board.test.ts` was already fixed; this PR handles the siblings detonating soonest.

## Changes

- **`dublin-hash.test.ts`** — the "filters events outside the date window" test pinned `16 March 2026` with `days:90` and asserted it was included (85 days old → fails ~June 14). Near-term row is now relative to `now` via a `relativeDate()` helper; the 2099/1990 rows stay static (always outside ±90).
- **`sdh3.test.ts`** — same ±90 bomb (`March 20, 2026` → ~June 18): made relative. Also widened the two `days:365` hareline fetch tests to an all-time window, since they reuse the date-pinned `HARELINE_HTML` parser fixture (which legitimately stays fixed for the parser-direct tests).
- **`squarespace-events.test.ts`** — densely coupled to fixed 2026 dates via `applyDateWindow`; froze the clock (Date-only) to `2026-06-01` in the fetch `describe`. No test there asserts now-relative exclusion, so a fixed reference date is safe and needs zero assertion changes. Date-only faking keeps `setTimeout` real so awaited fetch mocks still resolve.

## Verification

- All four files (incl. atlanta) pass now **and** with the clock advanced to `2027-06-09` — past every original detonation date.
- `npx tsc --noEmit` and `eslint` clean. Diff is test-only (3 files, +88/−11).

## Follow-up (not in this PR)

A clock-advanced sweep surfaced **~92 more tests across 32 adapter files** with the same pattern, detonating **2027–2028** (none imminent; meetup alone is 27). Catalogued with reproduction method and fix patterns in #2066 — left open intentionally as the backlog for that broader sweep.

### DIFF
--- a/src/adapters/html-scraper/dublin-hash.test.ts
+++ b/src/adapters/html-scraper/dublin-hash.test.ts
@@ -43,6 +43,29 @@ function mockFetchResponse(html: string) {
   } as Response);
 }
 
+// Build a date relative to "now" in both the parser's display format
+// ("D Month YYYY") and ISO ("YYYY-MM-DD"). The window-filter test below must
+// keep its "near-term" event inside buildDateWindow(90); pinning an absolute
+// date silently rots once the wall clock passes it (the same time-bomb that hit
+// atlanta-hash-board on 2026-06-08). Far-past/far-future rows can stay static —
+// 1990 and 2099 are outside ±90 days for any plausible run date.
+const MONTH_NAMES = [
+  "January", "February", "March", "April", "May", "June",
+  "July", "August", "September", "October", "November", "December",
+];
+function relativeDate(daysFromNow: number): { display: string; iso: string } {
+  const d = new Date();
+  d.setUTCHours(12, 0, 0, 0);
+  d.setUTCDate(d.getUTCDate() + daysFromNow);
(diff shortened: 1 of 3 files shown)


johnrclem:
@codex @claude review

johnrclem:
Addressed the `relativeDate` duplication nitpick (raised by both CodeRabbit and claude[bot]) in c7f9d1ab: extracted the helper to `src/test/date-helpers.ts` and updated both `dublin-hash.test.ts` and `sdh3.test.ts` to import it. The shared helper returns composable parts (iso/compact/weekday/monthName/day/year) so each test builds its own source-specific display string. tsc + lint clean; all four de-bombed files pass now and at a 2027 clock.

PR #2176, opened 2026-06-13
fix(safe-fetch): default timeout on direct-fetch branch (closes #2174)

johnrclem:
Mirrors the residential-proxy branch: one default 45s signal hoisted before the redirect loop bounds total request time across hops; caller signal still overrides. Adds safe-fetch.test.ts (3 tests).


johnrclem:
@claude @codex review

PR #2248, opened 2026-06-17
feat(merge): scrub mid-string phone/email PII from hares for every adapter

johnrclem:
## Context

The merge pipeline's `sanitizeHares` stripped only **trailing** logistics/boilerplate — it did **not** remove phone numbers or emails embedded mid-string in a hare line. Any source listing hare contacts inline therefore leaked PII into canonical, public-facing events. The Seoul H3 fix ([#2227](https://github.com/johnrclem/hashtracks-web/pull/2227)) was Korea-only and wired into a single adapter.

This generalizes phone/email scrubbing so **every adapter's `hares` is scrubbed at ingest**, and clears the existing prod backlog.

## Changes

- **New shared module** `src/adapters/hare-pii.ts` — `containsHarePii` / `scrubHarePii` / `HARE_PII_RES`, four anchored patterns kept as separate literals (Sonar S5843):
  - email
  - generic `+CC` international — digit-count floor `\+\d{1,3}[\s().-]*\d(?:[\s().-]*\d){6,13}` so the hash "+1" plus-one convention isn't clipped
  - global NA phone via `new RegExp(PHONE_NUMBER_RE.source, "g")` — single source of truth shared with the audit's `hare-phone-number` rule so `selfHealSanitizers` converges
  - Korean `\b01\d…` (load-bearing — NA's 3-3-4 / bare-10 doesn't match Korean 3-4-4)

  Detection uses `.search()` (avoids the `/g` `lastIndex` bug). `scrubHarePii` also has a regex-free paren-artifact cleanup so `"Just Jorge (973-760-5774)"` → `"Just Jorge"` instead of the dangling `"Just Jorge )"` the NA pattern's `\(?` would leave (balanced parens like `(1995-1996)` preserved).
- **Wired into `sanitizeHares`** — gated on `containsHarePii` so non-PII hares pass through **byte-identical** (no separator normalization → no fingerprint/audit churn). Runs before the boilerplate search so co-hares survive a mid-string phone (`"Alice (415) 555-1212 & Bob"` → `"Alice & Bob"`).
- **`sh3-pii.ts` reduced to a re-export shim** — Seoul adapter + its regression test unchanged (defense in depth, single source of truth).
- **One-off backfill** `scripts/backfill-hare-pii-scrub.ts` (dry-run default, `--apply` with optimistic-lock guard, scoped to PII-containing rows).

## Verification

- `npx tsc --noEmit` — clean
- `npm run lint` — 0 errors (the `new RegExp` suppression mirrors `hare-extraction.ts`)
- `npm test` — **9291 passed, 0 failures** (new `hare-pii` suite, extended `sanitizeHares` block incl. a gate-preservation test, Seoul regression intact)
- Backfill against prod: dry-run found **30 / 34,976** rows with PII → reviewed → `--apply` wrote 30 → re-run confirms **0 remaining** (idempotent)

## Known, intentional gaps

Two conservative misses (prod-verified as the only ones), both rows still had their other PII stripped so neither re-triggers detection:
- leading-`0` national-without-`+` numbers — a pattern for them would false-positive on dotted dates like `07.08.2024`
- obfuscated `" at "` emails — too false-positive-prone to catch

### DIFF
--- a/scripts/backfill-hare-pii-scrub.ts
+++ b/scripts/backfill-hare-pii-scrub.ts
@@ -1,0 +1,72 @@
+/**
+ * One-off full-history PII scrub of canonical Event.haresText (#2227 follow-up).
+ *
+ * The merge pipeline's `sanitizeHares` now strips phone numbers / emails embedded
+ * mid-string, so all FUTURE merges are protected. The daily audit's
+ * `selfHealSanitizers` also re-sanitizes the rolling −7d…+90d window. This script
+ * clears the historical backlog OUTSIDE that window in one shot.
+ *
+ * Scoped & safe:
+ *   - Only touches rows where `containsHarePii(haresText)` is true — non-PII hares
+ *     are never re-sanitized for unrelated reasons (no corpus churn).
+ *   - Computes the new value with the SAME `sanitizeHares` the pipeline uses.
+ *   - Optimistic-lock guard (`where: { id, haresText: <old> }`) so a concurrent
+ *     scrape can't be clobbered (mirrors selfHealSanitizers in audit-runner.ts).
+ *   - A null result clears the field (explicit clear).
+ *   - Idempotent: once scrubbed the row no longer matches `containsHarePii`.
+ *   - haresText is NOT part of the fingerprint, so rewriting it never disturbs dedup.
+ *
+ * Run (Railway proxy uses a self-signed cert):
+ *   Dry-run: set -a && source .env && set +a && BACKFILL_ALLOW_SELF_SIGNED_CERT=1 npx tsx scripts/backfill-hare-pii-scrub.ts
+ *   Apply:   BACKFILL_ALLOW_SELF_SIGNED_CERT=1 npx tsx scripts/backfill-hare-pii-scrub.ts --apply
+ *   Env:     DATABASE_URL
+ */
+import { containsHarePii } from "@/adapters/hare-pii";
+import { sanitizeHares } from "@/pipeline/merge";
+import { runOneShot } from "./lib/one-shot";
+
+const SAMPLE_LIMIT = 30;
+
+void runOneShot(async ({ prisma, apply }) => {
+  const events = await prisma.event.findMany({
+    where: { haresText: { not: null } },
+    select: { id: true, haresText: true },
+    orderBy: { date: "asc" },
+  });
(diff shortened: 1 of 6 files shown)


johnrclem:
@codex @claude review

johnrclem:
Addressed review feedback in `2f928e38`:

**@chatgpt-codex-connector P1 — no-space parenthesized NA phones.** Confirmed real: `PHONE_NUMBER_RE` requires a `[-.\s]` after the optional `\)?`, so `"Alice (415)555-1212"` was not detected and the scrub never ran. Added `PAREN_NA_PHONE_RE` (`\(\d{3}\)[-.\s]?\d{3}[-.\s]?\d{4}(?!\d)`) to `HARE_PII_RES` — it covers both `(415)555-1212` and `(415)5551212`. It's a scrub-side **superset** of the audit detector (scrub ⊇ detect), so `selfHealSanitizers` stays convergent. Tests added in both `hare-pii.test.ts` and the `sanitizeHares` block. Re-ran the prod backfill afterward → still **0 rows** (no events were in this format).

**@claude obs #1 — exported `/g` regex footgun.** Dropped the four individual regex exports (`EMAIL_RE`/`INTL_PHONE_RE`/`DOMESTIC_MOBILE_RE`/`NA_PHONE_GLOBAL_RE`) — nothing imported them, and exposing live `/g` objects invited a `.test()`/`.exec()` `lastIndex` bug. Only `containsHarePii`/`scrubHarePii`/`HARE_PII_RES` are public now.

**SonarCloud hotspot (S5852 ReDoS on `EMAIL_RE`, line 30).** Reviewed and marked **Safe**: the domain part `[\w-]+(?:\.[\w-]+)+` has a disjoint required `\.` separator between repetitions (the `[\w-]` label class excludes `.`), so there's no overlapping-quantifier ambiguity / super-linear backtracking — the textbook safe email-domain form — and it only runs on the ≤200-char capped `hares` string. Quality gate is now green.

**Acknowledged, no change:**
- `EMAIL_RE`'s `[\w.+-]` — the `-` is the last char in the class, so it's a literal hyphen, not a `+`→`-` range; `,` is **not** matched. (Verified.)
- semicolon→comma normalization only fires when PII is present (gated by `containsHarePii`); clean hares pass through byte-identical.
- backfill loads rows into memory — fine at 34,976×2 cols; one-shot, already run.

Issue #2238, opened 2026-06-17
Seletar H3 — backfill ~2,075 historical runs (1980–present) absent from HashTracks; HTML_SCRAPER only covers upcoming window


johnrclem:
**Finding:** The Seletar H3 source enumerates the kennel's complete run history back to run #1 (24 Jun 1980), but HashTracks only holds the recent upcoming window (~13 upcoming events). The bulk of the kennel's ~46-year history is missing.

**Source:** Seletar H3 PWA (HTML_SCRAPER) — https://sh3app.hash.org.sg/php/util/HashController.php
**Sample HashTracks event (current run #2384):** https://www.hashtracks.xyz/hareline/cmnpkxc42000m04l1h5dckm2m

**Source enumeration (run-level fields only):**
- Distinct runs: **2,088**, run numbers 1 → 2399, dates **1980-06-24 → 2026-10-06**.
- By decade: 1980s 226, 1990s 545, 2000s 533, 2010s 536, 2020s 248.
- Field coverage across those runs: date 2088/2088; run site 334; GPS 295; hares present 333; per-run comment/title 110; run length 0; run type 0.

**Why not a wider scrape:** This is an HTML_SCRAPER (partial enumeration) — widening the adapter window is unsafe because the reconcile step would cancel events the adapter doesn't return in a single pass. Per audit playbook, the safe path is a **one-shot DB insert** of the historical rows (run number, date, run site, GPS, hares, optional comment) without touching the adapter.

**Data access note:** the source view (`vw_hareline`) also exposes member PII (email, mobile, birthdate, photo paths). Any backfill must map **only** run-level fields + hare hash-names and must not import member contact/PII columns.

**Fix hypothesis:** Schedule a one-shot historical import job that pulls the full `vw_hareline` run list (run no, datetime, runsite, gps, hares hash-names, comment) and inserts past runs as completed events, leaving the live HTML_SCRAPER adapter untouched for the upcoming window.

johnrclem:
Backfilled via the refactored one-shot `scripts/backfill-seletar-h3-history.ts`.

[...]

Issue #2221, opened 2026-06-15
Samurai H3 — Missing event #2642 (Samurai Family H3, 19-Jul) present on source, absent from HashTracks


johnrclem:
The Samurai H3 hareline source lists run #2642 but HashTracks skips from #2641 to #2643.

**Source:** https://samuraihash2017.wixsite.com/samurai/hare-line (HTML_SCRAPER)

**Missing row (#2642):**
- Date: 19-Jul (venue marked T.B.D.)
- Time: 13:00
- Venue: (blank / T.B.D.)
- Train: (blank)
- Hare: (blank)
- Fee: 1000jpy Non-Alcohol; 500jpy kids free
- Note: Samurai Family H3
- Run #: 2642

HashTracks upcoming list goes #2641 (18-Jul) -> #2643 (25-Jul); #2642 is not present in upcoming or past.

**Path:** Source is HTML_SCRAPER (partial-enumeration: only shows a current/upcoming window starting 13-Jun). A wider scrape window is unsafe (reconcile would cancel rows the adapter didn't return). Propose a one-shot DB insert to add #2642 without touching the adapter. Available fields above (title=Samurai Family H3 special event, date 19-Jul, start 13:00, fee tiers, run #2642).

johnrclem:
Verified self-healed — no insert needed.

Prod already holds run #2642 as a CONFIRMED canonical Event:
- `#2642 2026-07-19 | CONFIRMED | canon=true | start=13:00 | Samurai H3 #2642`, sitting correctly between #2641 (2026-07-18) and #2643 (2026-07-25).

The live Samurai adapter now returns #2642 with full data (the source filled in the previously-TBD fields): hares "Kick My Balls", venue "Chitose-Funabashi Sta. Kasamori Park", `Fee: 1000jpy, Non-Alcohol 500jpy, kids free`, `Note: Samurai Family H3`. The original miss was a transient gap the daily scrape resolved.

A one-shot insert would have been a no-op (dedup against the existing event), so none was shipped — closing per the verify-before-acting path.

PR #2262, opened 2026-06-18
fix(seletar): backfill 1980→present history via merge pipeline + reconcile safety (#2238)

johnrclem:
## Summary

One-shot history backfills for two chrome-kennel audit issues. Scope was `scripts/**` only; the `sources.ts` one-liner was added after an adversarial review surfaced a reconcile hazard (see below).

### #2238 — Seletar H3: ~2,072 historical runs backfilled (1980→present)
The existing `scripts/backfill-seletar-h3-history.ts` inserted RawEvents directly with `processed: false` — which **never become canonical Events** (`scrapeSource` only merges the live adapter's fetch results, so pre-inserted raws stay orphaned; same trap the kljh3 refactor documents). Running it as-is would have created ~2,075 invisible orphan rows.

Refactored to the blessed kljh3 pattern (`runBackfillScript → reportAndApplyBackfill → processRawEvents`), reusing the adapter's shared `fetchSeletarRows` / `groupSeletarRows` / `HISTORICAL_SQL` exports (PII filtering intact). **Applied to prod:**
- `created=2061 updated=7 skipped=4 blocked=0 eventErrors=0`; idempotent re-run → `created=0 skipped=2072`.
- seletar-h3 now has **2,084 CONFIRMED canonical Events**, range **#1 (1980-06-24) → #2396 (2026-09-15)**.

### #2221 — Samurai H3 #2642: verified self-healed (no code)
Prod already holds `#2642 2026-07-19 | CONFIRMED | start=13:00`, correctly between #2641 and #2643, and the live adapter now returns it with full data (the source filled in the previously-TBD venue/hare). A one-shot insert would have been a redundant no-op, so the issue was closed with evidence rather than shipping a fragile artifact.

## Reconcile safety (from Codex adversarial review)
The Seletar recurring adapter is future-only (`UPCOMING_SQL = hl_datetime >= CURDATE()`), and the source lacked `upcomingOnly`, so the reconciler's `now ± scrapeDays` window would CANCEL the sole-source historical Events the backfill creates (future-only scrape never re-emits them; cancellation doesn't self-heal). Fixed:
- Added `config: { upcomingOnly: true }` to the Seletar source so reconcile only cancels future-dated orphans.
- Added a warn-only, apply-mode guard in the backfill script that flags when the live source row isn't yet `upcomingOnly`.
- Applied `upcomingOnly: true` to the **prod** source via a targeted update (not a full seed) — verified: 13 recent-window events CONFIRMED, **0 CANCELLED**.

## Verification
- `npx tsc --noEmit` ✓ · `npm run lint` ✓ (0 errors) · `npm test` ✓ (318 files / 9,343 tests)
- Prod queries confirm Seletar history present + Samurai #2642 present; both issues closed.

## Notes
- **Duplicate `sources.ts` edit:** a parallel session (started from a spawned chip) is also adding `upcomingOnly` to the Seletar source — same one-liner, same intent; trivial dedupe on merge.
- **Seed durability:** prod config is set now and stays durable once this merges + a normal `db seed` runs from updated main.

Issues: #2238, #2221 (both already closed with result summaries).


johnrclem:
@codex @claude review

Issue #2260, opened 2026-06-18
SFFMH3 — ~99 historical FMH3 events on sfh3.com not in HashTracks (HTML_SCRAPER partial-enumeration)


johnrclem:
## Summary
The SFH3 multihash sources map kennel code **FMH3 ("Fully Mooned H3 Crawls")** -> HashTracks **SFFMH3 (San Francisco Full Moon Hash)**. The mapping is confirmed working: the one upcoming FMH3 iCal event, "Flower Moon Pub Crawl" (May 8 2026), was merged correctly into SFFMH3 with full fields — https://www.hashtracks.xyz/hareline/cmoq4y7pg000804jv8vi4ny73 (title, hares=Mouth Down South, time 6:30-9:30 PM, location "Last Rites", rich description).

However the per-kennel **all-period** view exposes far more history that we are NOT capturing.

Source: https://www.sfh3.com/runs?kennel=7&kennels=7&period=all (FMH3 = kennel index 7)

## Counts / range
- Source lists **100 runs** spanning **2008–2026** (the table caps at 100 rows; no next-page link, so this is the visible ceiling).
- HashTracks SFFMH3 currently holds only **1** of these real pub-crawl events (Flower Moon Pub Crawl). The other ~99 historical FMH3 runs are absent. (The remaining 26 SFFMH3 events on HashTracks are STATIC_SCHEDULE lunar placeholders titled "SFFMH3 Full Moon Run", not these source rows.)

## Available fields per row (from the HTML table)
- date (When): all 100 rows
- title (What): ~60 rows have a specific title (e.g. "Cold Moon", "Pink Full Moon - an FMH3 takeover of Fog City", "Fully Mooned invasion of Surf City RDR")
- location (Where): ~53 rows
- hare (Hare): ~30 rows named (rest "Unknown")
- run number: only ~2 rows numbered; almost all are "(No #)"
- start time: NOT in the table; cost: not present

## Recommended path
This source is **HTML_SCRAPER (partial enumeration)** — the runs adapter only returns the upcoming window via /runs?kennels=all, so a widened scrape window is unsafe (reconcile would cancel the rows the adapter doesn't return). Per the deep-dive playbook, propose a **one-shot DB insert** of the ~99 historical FMH3 rows (date/title/location/hare as available) into SFFMH3 without touching the adapter. Run-number and start-time should be left null where the source omits them.

## Note
[...]

PR #2273, opened 2026-06-22
fix(admin-audit): persist deep-dive completions reliably + fail loud (#2261)

### LINKED ISSUE #1160: Admin audit — Mark deep dive complete dropdown is unsafe for Claude/Chrome automation (GGFM marked as Galveston)

### LINKED ISSUE #2261: Admin audit: 'Mark complete' for SFFMH3 did not register (kennel still in queue, coverage unchanged) — #1160
## What happened
On the admin audit Kennel Deep Dive panel, I set the Deep Dive Target to **SFFMH3 (San Francisco, CA)**, clicked **Mark complete**, confirmed the modal header read "Mark deep dive complete: SFFMH3 (San Francisco, CA)", entered Findings filed = 1 + a one-line summary, and clicked **Mark SFFMH3 complete**.

After a hard reload of /admin/audit:
- **SFFMH3 still appears in the QUEUE** with `last dived: never`.
- **Coverage is unchanged at 307 / 383** (same as before the submit).
- The Deep Dive Target reverted to the next-up queue kennel (Pinelake H3), as expected, but SFFMH3 was never recorded as dived.

This matches the misattribution / non-registration symptom tracked in **#1160**. Per the deep-dive runbook I stopped and did **not** re-submit (to avoid double-marking or marking the wrong kennel).

## Notes
- The actual audit finding for SFFMH3 was filed successfully via the audit API as **#2260** (historical FMH3 events backfill). Only the *completion bookkeeping* failed.
- No other queue kennel shows a recent `last dived` timestamp, so the submit appears to have silently no-op'd rather than landing on a visibly different kennel — but it should be checked server-side in case it was misattributed.

## Repro
1. /admin/audit → Kennel Deep Dive → select SFFMH3 in target dropdown.
2. Mark complete → set findings=1 → confirm.
3. Hard reload → SFFMH3 still in queue, coverage not incremented.

## Ask
Investigate why the deep-dive completion for SFFMH3 didn't persist, and whether it was misattributed to another kennel. Provide a way to manually record the SFFMH3 deep dive (findings filed = 1, see #2260) once root-caused.

johnrclem:
Fixes #2261.

## Problem
On `/admin/audit` → Kennel Deep Dive, marking **SFFMH3** complete silently no-op'd: after reload the kennel was still queued, coverage unchanged (307/383), **no `AuditLog` row written**, and the operator got no actionable signal. This is the silent-non-registration class tracked in #1160.

## Root cause
The deep-dive queue *and* its integrity snapshot are both capped at the top **20** kennels (`DEEP_DIVE_QUEUE_DEFAULT_LIMIT = 20`). On submit, `recordDeepDive` re-hashed the **entire** live top-20 and rejected with `queueChanged` if *any* kennel in that set had shifted — even though credit is already cryptographically pinned to the frozen `kennelCode`. On a live dashboard (daily 06:00 UTC cron ingest, parallel admin sessions, a 10-min token TTL spanning the operator's #2260 research) the top-20 churns routinely, so the single submit returned `{ok:false, error:"queueChanged"}` and the dialog printed *"…submit again to confirm"* — which the #1160 runbook forbids, stranding the dive as a permanent no-op. The snapshot-equality gate added **no** misattribution safety beyond the `kennelCode` binding.

**Not misattributed (Goal #1):** confirmed by read-only prod query — zero `KENNEL_DEEP_DIVE` rows for `sffmh3` and no stray credit elsewhere at the incident time. A row can only ever land for the verified, token-bound kennelCode.

## Fix
**Server (`src/app/admin/audit/actions.ts`)**
- Drop the `queueChanged` snapshot-equality rejection. Kept intact: token signature/expiry validation, the `kennelCode` binding (the real #1160 guard), and the `kennelGone` eligibility check. A legitimately-bound single submit now persists through benign queue churn.
- **Idempotent token consumption** (from the Codex adversarial review): a `KENNEL_DEEP_DIVE` row written at/after the token's mint time short-circuits to that row, so a retry / double-submit / replayed payload can't write duplicate rows or falsely advance `lastDeepDiveAt`. Closes #1160 ask #5. (Mint time derived as `expiresAt − QUEUE_TOKEN_TTL_MS`; no token-format change.)
- Add admin-only `recordDeepDiveManual` + `lookupKennelForDeepDive` for backfilling a missed completion (explicit, echoed kennelCode, no token) — deliberately in app code, not `scripts/**`.

**Client (`src/components/admin/AuditDashboard.tsx`)**
- `handleSubmit` now surfaces **every** failure loudly (exhaustive switch + catch-all); no silent `{ok:false}`.
- New *"Manually record a deep dive"* dialog that echoes the resolved kennel (`SFFMH3 — San Francisco, CA`) and keeps **Record** disabled until the echo matches the typed code — preserving the anti-misattribution property without a queue token.

## Recording the missed SFFMH3 dive
Open `/admin/audit` → "Manually record a deep dive", enter `sffmh3`, look up (confirms *SFFMH3 — San Francisco, CA*), set findings = 1, summary referencing #2260, Record. (The prod write is left to the operator as an outward-facing action.)

## Tests
- Failing→passing reproduction: queue-change-between-open-and-submit now persists (previously proved the no-op).
- Added coverage: expired token, `kennelGone`, **idempotent replay**, and the manual action (admin guard, negative findings, `kennelNotFound`, happy path, echo lookup).
- `npx tsc --noEmit` ✅ · `npm run lint` ✅ (0 errors) · `npm test` ✅ (**9448 passed**).

## Scope
Touches only `src/app/admin/audit/actions.ts`, its test, and `src/components/admin/AuditDashboard.tsx`. No `scripts/**`, `prisma/**` (incl. `schema.prisma`), or profile/hareline/logbook changes — no schema change needed.

### DIFF
--- a/src/app/admin/audit/actions.test.ts
+++ b/src/app/admin/audit/actions.test.ts
@@ -4,6 +4,7 @@ vi.mock("@/lib/auth", () => ({ getAdminUser: vi.fn() }));
 vi.mock("@/lib/db", () => ({
   prisma: {
     auditLog: {
+      findFirst: vi.fn(),
       findMany: vi.fn(),
       groupBy: vi.fn(),
       create: vi.fn(),
@@ -16,6 +17,7 @@ vi.mock("@/lib/db", () => ({
     },
     kennel: {
       findMany: vi.fn(),
+      findUnique: vi.fn(),
       count: vi.fn(),
     },
     auditIssue: {
@@ -74,6 +76,8 @@ import {
   getCloseReasonRatiosByStream,
   getDeepDiveQueueToken,
   recordDeepDive,
+  recordDeepDiveManual,
+  lookupKennelForDeepDive,
   getAuditSyncFreshness,
   resyncAuditIssues,
 } from "./actions";
@@ -384,12 +388,16 @@ describe("getDeepDiveCoverage", () => {
 
 describe("recordDeepDive", () => {
   const mockLogCreate = vi.mocked(prisma.auditLog.create);
+  const mockLogFindFirst = vi.mocked(prisma.auditLog.findFirst);
   const mockKennelFind = vi.mocked(prisma.kennel.findMany);
   const mockLogGroupBy = vi.mocked(prisma.auditLog.groupBy);
 
   const ORIGINAL_SECRET = process.env.AUDIT_QUEUE_TOKEN_SECRET;
   beforeEach(() => {
     process.env.AUDIT_QUEUE_TOKEN_SECRET = "test-secret-for-queue-tokens";
+    // No prior dive by default — the idempotency guard finds nothing
+    // and the create path runs. Replay tests override this.
+    mockLogFindFirst.mockResolvedValue(null as never);
     // Default queue: a stable list with NYCH3 present so the
     // happy-path test can mint and verify a real token. Tests that
     // exercise removed/changed-snapshot paths override.
@@ -512,10 +520,16 @@ describe("recordDeepDive", () => {
     expect(mockLogCreate).not.toHaveBeenCalled();
   });
 
-  it("returns queueChanged when the snapshot diverges (409-shape)", async () => {
-    // Token was minted with snapshot of [nych3, agnews], but the
-    // queue now also has philly-h3 added. NYCH3 still present, so
-    // we want a refresh-and-retry rather than a hard error.
+  it("persists the dive when the rest of the queue churned but the kennel is still present (#2261 — no false queueChanged)", async () => {
+    // Token was minted with snapshot of [nych3, agnews]; by submit
+    // time the live queue also has philly-h3 (a daily cron ingest or
+    // a parallel admin shifted the top-20). NYCH3 is still in the
+    // queue and the token is cryptographically bound to it, so the
+    // credit is unambiguous — the dive MUST persist rather than
+    // silently no-op. The full-queue snapshot adds no misattribution
+    // safety beyond the kennelCode binding, so its divergence here is
+    // benign.
+    mockLogCreate.mockResolvedValue({ id: "log_churn" } as never);
     mockKennelFind.mockResolvedValue([
       {
         kennelCode: "nych3",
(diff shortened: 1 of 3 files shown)


johnrclem:
@claude @codex review

PR #2541, opened 2026-07-04
feat(ical): maps.app.goo.gl support + documented deferred HashNYC iCal cutover

johnrclem:
## Context

hashnyc.com relaunched with a new site that publishes a clean structured iCal feed (`https://hashnyc.com/public/hareline.ics`), which would let us retire the bespoke `HashNYCAdapter` HTML scraper. During this work the site **reverted to the old HTML layout** (the `.ics` now 404s), so the **source cutover is deferred** — the old HTML scraper keeps running untouched — and only the safe, general adapter improvements land now. The full, verified cutover recipe is documented to resume when they re-launch.

## What's in this PR

**Shared iCal adapter — `maps.app.goo.gl` support** (`src/adapters/ical/adapter.ts`)
- `MAPS_URL_PATTERN` now recognizes the modern `maps.app.goo.gl` share-link host.
- A maps-shaped `URL:` property routes to `locationUrl` (preferred over the imprecise name-search fallback) instead of leaking into `sourceUrl`. Real event-page URLs (e.g. EBH3's `ebh3.com/runs/…`) don't match, so they're unaffected. General improvement — benefits any iCal feed.

**Forward-looking test coverage** (`src/adapters/ical/adapter.test.ts`)
- A `HashNYC` describe block with a fixture mirroring the real feed: guards the multi-kennel routing (incl. the tricky `nawwh3`-vs-`nah3` ordering) and the new map-link support. Runs and passes today; annotated as coverage for the deferred cutover.

**Deferred cutover docs** (`docs/hashnyc-ical-cutover.md`, indexed)
- Trigger condition, exact seed entries, and the **atomic** provision-and-retire migration (create new ICAL_FEED source + 12 SourceKennel links + disable old row in one transaction — no deploy-time gap). Verified end-to-end against a local prod-copy DB (applies, idempotent, correct routing incl. the JSONB `\b`-escaping round-trip). The atomic design came out of an adversarial review that caught the "disable without provisioning" gap.

**Drive-by fix** (`src/adapters/hashrego/adapter.test.ts`)
- An unrelated, pre-existing test failure: a hardcoded `2026-06-27` fixture date aged out of the forward scrape window (fails on clean `main` as of 2026-07-04). Made the row date relative to now so it can't age out again.

## Verification
- `npx tsc --noEmit` — 0 errors
- `npm run lint` — 0 errors (22 pre-existing warnings)
- `npm test` — 334 files, 9,859 tests pass
- iCal adapter live-verified against the feed while it was up (53 events, 0 unrouted, 8 with `maps.app.goo.gl` pins, 0 leaking into `sourceUrl`)

### DIFF
--- a/.claude/rules/documentation-index.md
+++ b/.claude/rules/documentation-index.md
@@ -17,5 +17,6 @@ globs:
 - `docs/test-coverage-analysis.md` -- Test coverage gap analysis and priorities
 - `docs/self-healing-automation-plan.md` -- Self-healing automation loop architecture, confidence scoring, roadmap
 - `docs/residential-proxy-spec.md` -- Architecture and deployment guide for residential proxy
+- `docs/hashnyc-ical-cutover.md` -- DEFERRED cutover: HashNYC HTML scraper -> iCal feed (ready-to-apply seed + migration; resume when hashnyc.com re-launches the .ics)
 - `infra/proxy-relay/` -- NAS-deployed residential proxy (Cloudflare Tunnel + Node.js forwarder)
 - `infra/browser-render/` -- NAS-hosted Playwright rendering service (Dockerfile + server.js)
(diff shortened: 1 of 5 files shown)


johnrclem:
@codex @claude review

PR #2654, opened 2026-07-15
feat(hc-batch-6): onboard Rio H3, Poznan H3, Divahhh, Steel City H3, Manchester H3, Douliu H3

johnrclem:
Implements the **six config-only `HARRIER_CENTRAL` handoffs** from the 2026-07-10→15 daily runs, which were generated but never implemented. Shipped as one batch PR per the hc-batch-4 ([#2622](https://github.com/johnrclem/hashtracks-web/pull/2622)) / hc-batch-10 ([#2526](https://github.com/johnrclem/hashtracks-web/pull/2526)) precedent. **Zero new adapter code.**

## 🔴 Why these were invisible

All six handoffs — plus their run-log entries, 4 platform notes and queue rows — sat **untracked/uncommitted in the main repo working tree**. `origin/main` never moved, so a session branching from `origin/main` saw a handoffs folder that looked complete, and an audit confidently concluded *"0 kennels left to onboard"* while six live kennels sat unimplemented on disk.

This is the **same failure the 2026-07-05 catch-up diagnosed** (`copy-newest-handoff.sh` only surfaces the *newest* handoff), whose recommended fix was never built. It has now recurred at 6× scale. The handoffs README gains an audit snippet + the root cause: **the daily run must commit its output.**

## The 6 kennels

| Kennel | code / slug | region.ts | `upcomingOnly` | Backfill | Live-verify |
|---|---|---|---|---|---|
| Rio H3 | `rioh3` / `rio-h3` | 3 (new Rio METRO) | ✅ | 3 (#444–446) | 0 upcoming (expected) |
| Poznan H3 | `ph3-pl` / `poznan-h3` | 3 (new Poznan METRO) | ✅ | 3 (#11–13) | 0 upcoming (expected) |
| Divahhh | `divahhh` / `divahhh` | 0 (Brussels exists) | ✅ | 9 (#36–44) | 0 upcoming (expected) |
| Steel City H3 | `sch3-gb` / `steel-city-h3` | 2 (new Sheffield METRO) | omitted | none (live-only) | **1** (#22 2026-07-18 12:00) |
| Manchester H3 | `mh3-gb` / `manchester-h3` | 2 (new Manchester METRO) | omitted | none (live-only) | **3** |
| Douliu H3 | `douliu-h3` / `douliu-h3` | 2 (new Douliu METRO) | ✅ | **31** | 0 upcoming (expected) |

[...]

PR #2655, opened 2026-07-17
fix(seed): guard prisma/seed.ts main() behind an entry-point check

johnrclem:
## The bug

`prisma/seed.ts` ends with an **unguarded top-level `main().catch(...)`**. So *any* `import { seedKennels, toSlug } from "./prisma/seed"` fires the **entire full seed** as an import side effect — `seedKennels` over all ~515 kennels + ~443 sources (overwriting profile fields + source config) **plus** the global `runScheduleRuleBackfill` (which runs `deactivateStaleRules`). This is exactly the full seed the post-merge runbook forbids (barbados #2471).

## How it bit us

During the hc-batch-6 post-merge ([#2654](https://github.com/johnrclem/hashtracks-web/pull/2654)), a scoped-seed one-shot imported `seedKennels` to reuse it with filtered arrays — and unknowingly ran a **full seed concurrently** against prod. The tell was a run of misleading `P2002` "unique constraint on (name)" errors on region names that **didn't exist** — two racing `region.create()`s from the two `main()`s. It looked like a driver / prepared-statement bug and cost real time to chase before the double `main()` was spotted.

Blast radius stayed **near-zero** only by luck: the one-shot's `process.exit(0)` killed the runaway `main()` right after region creation, before the table-wide overwrite stages. Verified against prod afterward — only the 6 batch kennels touched, zero other source configs changed, `deactivateStaleRules` never ran.

## The fix

Guard `main()` behind the same entry-point check `scripts/backfill-schedule-rules.ts` **already uses**, so importing `seed.ts` is inert and only `npx prisma db seed` / direct `npx tsx prisma/seed.ts` runs it:

```ts
const entryPoint = (process.argv[1] ?? "").replace(/\\/g, "/");
if (entryPoint.endsWith("/seed.ts") || entryPoint.endsWith("prisma/seed.ts")) {
  main().catch((e) => { console.error(e); process.exit(1); });
}
```

The Prisma seed command is `npx tsx prisma/seed.ts` (`prisma.config.ts`), so `process.argv[1]` is the absolute path ending `/seed.ts` — the guard passes for direct execution.

## Verification

- **Import is now inert:** `import { seedKennels, toSlug } from "./prisma/seed"` runs no seed output / no DB connection; `toSlug` still exports (`seedKennels` remains intentionally un-exported).
- **Direct execution still works:** `npx tsx prisma/seed.ts` against the local dev DB runs the full seed to `Seed complete!`.
- `npx tsc --noEmit` → 0 errors · `npm run lint` → 0 errors (22 pre-existing warnings, none here) · `npm test` → **9998 passed**.
- No file in the repo imports `prisma/seed` expecting the `main()` side effect (grep-confirmed).

## Docs (in this PR)

- `run-log.md`: **POST-MERGE LIVE** record for hc-batch-6 (6 kennels live, 46 backfill events, 0 cancelled on every source, scoped Pass 3) + the seed.ts import incident.
- `2026-07-15-hc-batch-6-retro.md`: a post-merge-incident section documenting the trap, the fix, and the runbook rule (never `import` from `prisma/seed.ts`; use self-contained upserts or the guarded `scripts/backfill-schedule-rules.ts`).

### DIFF
--- a/docs/kennel-onboarding/handoffs/retros/2026-07-15-hc-batch-6-retro.md
+++ b/docs/kennel-onboarding/handoffs/retros/2026-07-15-hc-batch-6-retro.md
@@ -208,6 +208,30 @@ await applyUpserts(prisma, planned);                 // NOT runScheduleRuleBackf
    one to the HC config would let live + backfill share a single rule. Blast radius today is one
    kennel, so this is a follow-up, not a blocker.
 
+## Post-merge incident — importing `prisma/seed.ts` auto-runs the FULL seed
+
+The post-merge runbook says "targeted `seedKennels`". Reusing the *real* tested `seedKennels` means
+importing it — but **`prisma/seed.ts` has an unguarded top-level `main().catch()`**, so
+`import { seedKennels } from "./prisma/seed"` fired the entire full seed as an import side effect,
+running concurrently with the scoped call against prod. The tell was a string of misleading `P2002`
+"unique on (name)" errors on region names that didn't exist (two racing `create`s) — which looked like
+a driver / prepared-statement bug and cost real time to chase before the double `main()` was spotted.
+
+The scoped one-shot's `process.exit(0)` killed the runaway `main()` right after region creation, so the
+damage was near-zero (only the 6 batch kennels touched; no other configs; `deactivateStaleRules` never
+ran). But the trap is real and will recur.
+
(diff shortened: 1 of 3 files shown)


johnrclem:
@claude @codex review

